PR #57, opened 2017-09-11
Password on prompt

amolthacker:
PR with a minor change to prompt for user password instead of providing it in plain text ... is this needed ?


dtde/spark/cli/cluster_create.py
```diff
@@ -18,14 +24,15 @@ def setup_parser(parser: argparse.ArgumentParser):
...
+    parser.add_argument('--wait', dest='wait', action='store_true')
     parser.add_argument('--username',
                         help='Username to access your cluster (required: --wait flag)')
-    parser.add_argument('--password',
-                        help="The password to access your spark cluster's head node. If not provided will use ssh public key.")
     parser.add_argument('--ssh-key',
                         help="The ssh public key to access your spark cluster\'s head node. You can also set the ssh-key in the configuration file.")
-    parser.add_argument('--no-wait', dest='wait', action='store_false')
-    parser.add_argument('--wait', dest='wait', action='store_true')
+    # See http://bugs.python.org/issue9338
```
timotheeguerin: not sure I get how this bug is related here

dtde/spark/cli/cluster_create.py
```diff
@@ -18,14 +24,15 @@ def setup_parser(parser: argparse.ArgumentParser):
                         help='VM size for nodes in your cluster')
     parser.add_argument('--custom-script',
                         help='Absolute path of custom bash script (.sh) to run on each node')
+    parser.add_argument('--no-wait', dest='wait', action='store_false')
```
timotheeguerin: any reason you move that above?

dtde/spark/cli/cluster_create.py
```diff
@@ -40,7 +47,7 @@ def execute(args: typing.NamedTuple):
     log.info("wait for cluster:        %s", args.wait)
     log.info("username:                %s", args.username)
     if args.password:
-        log.info("Password: %s", '*' * len(args.password))
+        log.info("Password: %s", '*****')
```
timotheeguerin: cool meant to do that

timotheeguerin:
I think we should push people to use ssh key more than password but this will make it better anyway. We should change the add-user also to stay consistent

amolthacker:
good point @timotheeguerin ... i went back and checked ... I thought the bug was relevant for a couple of test cases .. made a few changes and as suggested have this applicable for add_user too

dtde/spark/cli/cluster_create.py
jiata: nitpick - can we add some spaces after the 'Password:' str and before '%s' so that it lines up with the rest of the lines.

dtde/spark/cli/cluster_add_user.py
```diff
@@ -10,7 +9,7 @@ def setup_parser(parser: argparse.ArgumentParser):
                         help='The usernameto access your spark cluster\'s head node')
 
     auth_group = parser.add_mutually_exclusive_group()
-    auth_group.add_argument('-p', '--password',
+    auth_group.add_argument('-p', '--password', action=cli.Password, nargs='?', dest='password',
```
timotheeguerin: I think doing it this way is breaking the usage of a ssh public key as it will ask for it all the time even if you have ssh_key passed as an argument or in the secrets.yml

jafreck:
Fixed by #116

PR #58, opened 2017-09-11
Handle error on invalid batch service parameters

amolthacker:
PR to print Batch Service error if any of the input parameters (especially vm_size) are invalid

Pretty sure this can be done better …
-	Instead of handling exception in just one place, it can be probably be done everywhere the lib uses batch service client
-	Instead of relying on Batch Service to throw an error on invalid parameters should there be a validator for things like cluster-id string, job name, valid vm_size for the batch account in use and checks for ensuring Standard_A0, Standard_GS, Standard_DS/DSv2 are not used to tell the user which param is invalid ?


timotheeguerin:
yeah here I wouldn't catch the error at this point as we then lose it if using the clusterlib programatically(and it also kills the app). 

We should probably have it catch in the cli entry point and then catch and rethrow with better details in the lib

amolthacker:
yes makes sense ... guess this would be needed for a few different error points ...

jafreck:
Handled by #85

PR #60, opened 2017-09-11
Feature: forward different ports for ssh

### LINKED ISSUE #59: Support forwarding custom ports
Only spark and Jupyter are being forwarded for now.

timotheeguerin:
fix #59


timotheeguerin:
Updated docs

timotheeguerin:
actually I just thought of it but those port you forward to the node but not to the container so it doesn't do much I feel

paselem:
@jiata and @timotheeguerin - the design for this will also play with the public IP stuff since a user can/will specify ports there to expose "publicly". Going forward, any SSH or Port features should be considered for both SSH and Public IP

docs/10-clusters.md
```diff
@@ -106,7 +106,15 @@ Spark Jobs UI and the Jupyter notebook in the cluster:
...
-_Additional ports can be forwarded using standard ssh port forwarding syntax._
+You can also forward any other ports the same way you would do with ssh
+
+```sh
+# This will forward the remote port 5678 to the local port 1234
+azb spark cluster ssh --id spark --username admin -L 1234:localhost:5678
+
+# You can use this option multiple time to forward multiple ports
+azb spark cluster ssh --id spark --username admin -L 1234:localhost:5678  -L 1235:localhost:5679
+```
```
paselem: We may want to add a note here that additional ports should be available at cluster create time via the container since that needs to happen first. Perhaps just a link to that doc is sufficient.

PR #65, opened 2017-09-13
Feature: config files

### LINKED ISSUE #66: Provide config files for reusing default for the cluster settings

jafreck:
fix #66


dtde/config.py
```diff
@@ -26,3 +27,45 @@ def get() -> configparser.ConfigParser:
     if not global_config:
         load_config()
     return global_config
+
+
+class ConfigObject:
```
timotheeguerin: I think we should have 1 object like that per config file. So maybe rename it to be ClusterConfig. Might also be worth it to make a parent class ConifgContainer that handle the reading of a file from yaml to an dict

dtde/spark/cli/cluster_create.py
```diff
@@ -30,29 +30,52 @@ def setup_parser(parser: argparse.ArgumentParser):
 
 
 def execute(args: typing.NamedTuple):
+    # read cluster.yaml configuartion file, overwrite values with args
+    confg_obj = ConfigObject()
+    confg_obj.read_config_file()
+
+    if args.cluster_id is not None:
```
timotheeguerin: hhm, here I would have something more in the lines of config_obj.merge(
size=args.size,
...
)

dtde/spark/cli/cluster_create.py
```diff
@@ -30,29 +30,52 @@ def setup_parser(parser: argparse.ArgumentParser):
 
 
 def execute(args: typing.NamedTuple):
+    # read cluster.yaml configuartion file, overwrite values with args
+    confg_obj = ConfigObject()
```
timotheeguerin: rename cluster_config

.gitignore
```diff
@@ -1,4 +1,5 @@
 # custom
+cluster.yaml
```
timotheeguerin: shouldn't we move that to be in the config folder. Related to that I don't think it should be called config as this might conflict with the user project structure. `.thunderbolt` or `.thunderbolt-config` @paselem and @jiata what do you think?

config/cluster.yaml.template
```diff
@@ -0,0 +1,11 @@
+# base cluster settings
+base:
+    cluster_id: <id>
```
timotheeguerin: all of this is in the cluster file not sure if they all need to be prefixed with cluster. also the base key is probably not needed

.gitignore
jafreck: yeah, this is a mistake. The cluster.yaml and secrets.cfg should be moved into the config/ directory. `.thunderbolt-conf` is probably a better name as well.

timotheeguerin: @jiata seems to prefer `.thunderbolt`

config/cluster.yaml.template
```diff
@@ -0,0 +1,11 @@
+# base cluster settings
```
jiata: can this be flattened? the yaml parser will probably return something that begins with 'base', but i dont think we need that

config/cluster.yaml.template
jiata: i am of the opinion that you should not be able to specify cluster_id here (or at least not recommend setting it here)... 

Maybe, in the template, we can just comment it out and add a note saying "you can optionally choose to specify your cluster id here. However, please note that cluster ids need to be unique."

config/cluster.yaml.template
```diff
@@ -0,0 +1,11 @@
...
+base:
+    cluster_id: <id>
+    cluster_vm_size: <vm-size>
+    cluster_size: <number>
+    cluster_size_low_pri: <number>
+    cluster_username: <username>
+    cluster_password: <password>
+    cluster_ssh_key: <path-to-ssh-key>
+    custom_script: <path to custom script to run on each node> (optional)
+    wait: <true/false>
```
jiata: in general, we should be adding more comments to this config file. 
some points:
- why is there both ssh_key and password?
- can i really define both size_low_pri and size? (how do we represent mutually exclusive params?)
- how do i know which vm sizes are available to me? (good place to provide a link to the docs)

Even really obvious things:
- what is cluster_size? (is it number of nodes? number of executors? number of cores?)
- what is low-pri? (lets provide a link out)
- what does 'wait' do?

docs/10-clusters.md
```diff
@@ -27,6 +27,32 @@ NOTE: The cluster id (--id) can only contain alphanumeric characters including h
 #### Low priority nodes
 You can create your cluster with [low-priority](https://docs.microsoft.com/en-us/azure/batch/batch-low-pri-vms) VMs at an 80% discount by using **--size-low-pri** instead of **--size**. Note that these are great for experimental use, but can be taken away at any time. We recommend against this option when doing long running jobs or for critical workloads.
 
+#### Configuration
```
jiata: i dont 100% that these docs should go here. I think we should make a **docs/11-cluster-config.md**

The cluster config stuff is for a slightly different use case. Also i think that this cluster config section will be large enough eventually (with all the other config work) that it will be warrant having its own page just for space.

docs/10-clusters.md
```diff
@@ -27,6 +27,32 @@ NOTE: The cluster id (--id) can only contain alphanumeric characters including h
...
+Set your cluster's configuration in a single location using the config/cluster.yaml file. A template with all available configuration settings is provided at config/cluster.yaml.template.
+
+For example, with the following configuration:
```
jiata: @timotheeguerin - we chatted about having a command that will create the config folder (.thunderbolt) - are we still doing that for this PR? 

What is the command? azb spark cluster generate-config

BTW
I'm thinking that we should call it .thunderbolt-cluster now because eventually we'll have a .thunderbolt-job right?
So we'll potentially also need a: azb spark job generate-config

dtde/config.py
```diff
@@ -26,3 +27,45 @@ def get() -> configparser.ConfigParser:
...
+                    self.cluster_size = 0
+                self.cluster_size_low_pri = config['base']['cluster_size_low_pri']
+                if self.cluster_size_low_pri is None:
+                    self.cluster_size_low_pri = 0
+                self.cluster_username = config['base']['cluster_username']
+                self.cluster_password = config['base']['cluster_password']
+                self.cluster_ssh_key = config['base']['cluster_ssh_key']
+                self.custom_script = config['base']['custom_script']
+                self.wait = config['base']['wait']
+            except ymal.YAMLError as err:
```
jiata: we should probably do more error handling here right?

if you specify both size-low-pri and size, we should stop the process and spit out a msg saying that it failed

requirements.txt
```diff
@@ -2,3 +2,4 @@ azure-batch==3.0.0
 azure-storage==0.33.0
 pytest==3.1.3
 pylint==1.7.2
+pyyaml==3.12
```
paselem: Just because we may need to think about this (way?) down the road, should we have a pattern where we split this file up into development and runtime sections?

```text
# production
azure-storage==0.33.0
...

# development
pytest=...
pylint=...
```

Honestly I have no clue if this is even supported in requirements files though...

timotheeguerin: yeah you can add comments in the file https://github.com/pypa/pip/issues/564

config/cluster.yaml.template
paselem: Hrm... maybe... I'm not sure what the _right_ thing to do here is. It makes the interactive mode much more straight foward.

We spoke about this yesterday as well - do we need a .template file for this? It feels redundant if we ship a fully fleshed out file, which makes getting started far easier.

.gitignore
```diff
@@ -1,4 +1,5 @@
...
 secrets.cfg
```
paselem: Can we remove the secrets.cfg as part of this?

timotheeguerin: ? secrets is the most important one to be here

dtde/config.py
paselem: +1. same goes for password and ssh_key. Oh, and also if they want a username, should we warn them if WAIT == false? (that last one should be a no-op once the service supports this feature for us)

.gitignore
paselem: Sorry, I meant to imply that I thought this was going to change to secrets.yaml, but I wasn't sure if it was part of this larger change.

config/cluster.yaml.template
paselem: This is similar to what do we do doAzureParallel with the generateConfigTemplate. We fill out a bunch of default values already. Actually, for doAzPar we should jsut ship a default template too :)

jiata: i **think** i agree with this :P 

definitely should be using a .template for the spark config files (and maybe even the ssh.yaml one too)

paselem: I think what I am saying to NOT have a .template for this one, just a cluster.yaml which looks like

```yaml
base:
    id: Spark
    vm_size: Standard_a2
    size: 0
    size_low_pri: 3
    ...
```

config/secrets.yaml.template
```diff
@@ -0,0 +1,14 @@
+batch:
```
jiata: can we add a comment at the top of this file that links back to the docs to show people who to create their batch/storage accounts

docs/10-clusters.md
```diff
@@ -27,6 +27,32 @@ NOTE: The cluster id (--id) can only contain alphanumeric characters including h
...
+
+```yaml
+# base cluster settings
+base:
```
jiata: this is not updated right?

also, this section is redundant now that you have your new 12-configuration.md file right?

config/cluster.yaml
```diff
@@ -0,0 +1,14 @@
+##  cluster settings
+# id: <id of the cluster to be created>
+# vm_size: <vm-size, see available options here: https://azure.microsoft.com/en-us/pricing/details/virtual-machines/linux/>
+# size: <number of dedicated nodes in the cluster, not that clusters must contain all dedicated or all low priority nodes>
+# size_low_pri: <number of low priority nodes in the cluster, mutually exclusive with size setting>
+# username: <username for the linux user to be created> (optional)
+# custom_script: <path to custom script to run on each node> (optional)
```
jiata: we should add docker-repo as a variable too. we can do "azb spark cluster create ... --docker-reop <repo>"

config/cluster.yaml.template
jiata: agree - but without 'base'

docs/10-clusters.md
jafreck: Yes, this is redundant and should be removed.

config/cluster.yaml
jafreck: Is that in the scope of this PR?

jiata: good point, might not be. or might want to do it when you merge. not sure what the best practice for this is.

config/cluster.yaml
```diff
@@ -0,0 +1,14 @@
+##  cluster settings
+# id: <id of the cluster to be created>
```
timotheeguerin: I would show the documentation above each entry instead of at the top
```yaml
# Id of the cluster to be created
id: spark_cluster

# VM Size, see available options here: https://azure.microsoft.com/en-us/pricing/details/virtual-machines/linux
vm_size: standard_a2
```

config/cluster.yaml
```diff
@@ -0,0 +1,14 @@
...
+# vm_size: <vm-size, see available options here: https://azure.microsoft.com/en-us/pricing/details/virtual-machines/linux/>
+# size: <number of dedicated nodes in the cluster, not that clusters must contain all dedicated or all low priority nodes>
+# size_low_pri: <number of low priority nodes in the cluster, mutually exclusive with size setting>
+# username: <username for the linux user to be created> (optional)
+# custom_script: <path to custom script to run on each node> (optional)
+# wait: <true/false> 
+
+id: my_spark_cluster
```
timotheeguerin: probably just spark_cluster so it looks better if they don't change it

.gitignore
```diff
@@ -1,6 +1,7 @@
 # custom
-secrets.cfg
+secrets.yaml
```
timotheeguerin: isn't that part of the `.thunderbolt` folder?

dtde/config.py
```diff
@@ -1,28 +1,144 @@
...
+            raise Exception(
+                    "Configuration file doesn't exist at {0}".format(path))
+
+        with open(path, 'r') as stream:
+            config = yaml.load(stream)
+            if config is None:
+               return
+
+            try:
+                if 'id' in config:
```
timotheeguerin: can you move all this section into a method `merge_dict` or something like that

dtde/config.py
```diff
@@ -39,21 +35,153 @@ def cleanup_spark_config():
...
+                print(type(self.wait))
+
+        except yaml.YAMLError as err:
+            print(err)
+
+    def merge(self, uid, username, size, size_low_pri, vm_size, ssh_key, password, wait, custom_script, docker_repo):
+        """
+            Merges fields with args object
+        """
+        if self.uid is None and uid is None:
```
timotheeguerin: can this merge use the _merge_dict instead of repeating most of the stuff?

PR #68, opened 2017-09-14
Feature: custom docker images

jiata:
#61 

- azb spark cluster create ... --docker-repo some-user/some-repo:tags
- using pyenv to manage python versioning
- consolidated dockerfiles to just be one where you pass in parameters at build-time to specify spark and python version
- documentation on building your own docker images
- symlink for $SPARK_HOME to always point to /home/spark-current (as opposed to something like /home/spark-2.1.0-bin-hadoop2.7)

Other things that should be address as separate PRs:
- remove jupyter install from node_scripts and set it as a optional (but highly recommended) custom-script or plugin of sorts
- docker auth
- migrating off 'jiata/spark' to something not personal to me
- i believe that our Dockerfile can be trimmed down a bit
- preloaded jupyter sample code does not work for spark 1.6.3 - in the sample code, we call "spark" which automatically fetches the Spark Session, but Spark Sessions do not exist in Spark 1.6.3! We have to call "SparkContext" instead

Misc:
- folks using Thunderbolt version of milestone <= 0.1.0 will still remain dependent on jiata/spark:latest - Currently, the default docker repo is pointing at jiata/spark:2.2.0-python3.5.4 but should be pointing at jiata/spark:latest. however, i cannot overwrite jiata/spark:latest as previous versions depend on that hosted image.


docker-image/Dockerfile
```diff
@@ -0,0 +1,66 @@
+# Ubuntu 16.04 (Xenial)
+FROM ubuntu:16.04
+
+# set version of python required for thunderbolt application
+ENV THUNDERBOLT_PYTHON_VERSION=3.6.2
+
+ARG SPARK_VERSION_KEY=spark-2.1.0-bin-hadoop2.7
+ARG PYTHON_VERSION=3.6.2
```
timotheeguerin: Maybe add a comment here about what is this version and what modifiying do

node_scripts/install/spark.py
```diff
@@ -63,17 +64,17 @@ def generate_jupyter_config():
...
             "ipykernel",
             "-f",
             "{connection_file}",
         ],
         env=dict(
             SPARK_HOME=spark_home,
-            PYSPARK_PYTHON="/usr/bin/python3",
-            PYSPARK_SUBMIT_ARGS="--master spark://{0}:7077 pyspark-shell".format(
-                master_node_ip),
+            PYSPARK_PYTHON="python",
```
timotheeguerin: so wait is it using the docker file PYSPARK_PYTHON env at all. we should be able to remove one or the other

jiata: oh yeah. they are the same thing actually... so i should just just the env var probably

jiata: actually, i take back this comment - i think we should actually just python instead of the env var... because if you select python3.5.4 on your docker image, you would pretty much expect that jupyter will also run with pyspark python version 3.5.4... i can't imagine a case where that isn't true.

PR #77, opened 2017-09-18
Feature: ssh configuration file support

### LINKED ISSUE #63: rename --webui to --jobui and rename --masterui to --webui in cli

### LINKED ISSUE #75: .ssh config file

jafreck:
Fix #75, fix #63


config/ssh.yaml
```diff
@@ -0,0 +1,19 @@
+# ssh configuration
+
+# cluster_id: <id of the cluster to connect to>
+cluster_id: spark_cluster
```
paselem: Now that I'm looking at this, I'm not sure if we should be adding the cluster id to this file. The ssh should be valid for any ssh session to any cluster (and we should expect that users create and throw clusters away often) so adding the cluster id in here seems to go against that principle. 

Thoughts?

jiata: adding cluster id has always been weird for me. for ssh and for create cluster. i think id should be mandatory param for both cluster creation and cluster ssh

for spark cluster create:
The scenario of someone getting an error because of duplicate cluster name is too likely. For simplicity, i think we should just make it a standard that id needs to be specified for each of the azb spark cluster xyz commands.

for azb spark cluster ssh - prefilling the cluster_id in the configs seems like not a good idea i think.. because how do we know that the user is gonna have a cluster called "spark_cluster"

jafreck: I agree that errors because of duplicate clusters is very likely with how things are currently set up. Moving id to a parameter seems like a fine solution.

The reason that the default ssh.yaml file has spark_cluster as the cluster_id is because the same id is used as default in cluster.yaml. This is so users could type `azb spark cluster create; azb spark cluster ssh` and it would just work.

config/ssh.yaml
```diff
@@ -0,0 +1,18 @@
+# ssh configuration
+
+# cluster_id: <id of the cluster to connect to, reccommended to specify with --id command line parameter>
+
+# username: <name of the user account to ssh into>
+username: spark
+
+# job_ui_port: <local port where the job ui is forwarded to>
+job_ui_port: 8080
```
jiata: i think we should just stick with the standard jobui=4040, webui=8080 and jupyter=8888

those would be the ports that users expect.

docs/12-configuration.md
```diff
@@ -21,7 +20,34 @@ running `azb spark cluster create` will create a cluster of 4 Standard\_A2 nodes
...
+job_ui_port: 8080
+
+# web_ui_port: <local port where the spark master web ui is forwarded to>
+web_ui_port: 8081
+
+# jupyter_port: <local port which where jupyter is forwarded to>
+jupyter_port: 8082
+```
+
+running the command `azb spark cluster ssh` will attempt to ssh into a cluster called spark_cluster with the username spark. It will forward the job web ui to localhost:8080, the spark master web ui to localhost:8081 and jupyter to localhost:8082.
```
jiata: some nitpicks:

- "Running..."(capitalize)
- put names like 'spark_cluster' and the username 'spark' in quotations
- Spark Job UI (not job web ui)
- Spark should be capitalized and UI to
- Jupyter should be capitalized

jafreck: Yeah, this was written pretty lazily. Needs to be updated to reflect taking out the cluster_id as a default, too.

config/ssh.yaml
```diff
@@ -0,0 +1,18 @@
...
+username: spark
+
+# job_ui_port: <local port where the job ui is forwarded to>
+job_ui_port: 4040
+
+# web_ui_port: <local port where the spark master web ui is forwarded to>
+web_ui_port: 8080
+
+# jupyter_port: <local port which where jupyter is forwarded to>
+jupyter_port: 8088
```
jiata: this should be 8888

PR #79, opened 2017-09-18
Feature: added azb spark init command

### LINKED ISSUE #74: azure spark init

jafreck:
Fix #74


docs/00-getting-started.md
```diff
@@ -53,15 +59,17 @@ Once you have one, simply log in and go to the [Azure Portal](https://portal.azu
 
 4. Save your account credentials into the secrets.yaml file
 
-- Copy the secrets.yaml.template file to secrests.cfg
+- Copy the secrets.yaml.template file in the .thunderbolt/ directory to .thunderbolt/secrets.yaml
```
timotheeguerin: remove this?

docs/00-getting-started.md
```diff
@@ -53,15 +59,17 @@ Once you have one, simply log in and go to the [Azure Portal](https://portal.azu
...
 
 Windows
 ```sh
+cd .thunderbolt
```
timotheeguerin: I think the azb spark init should also rename that file for you and later(other pr) we can have it ask for the fields as they run `azb init`

jafreck: good point, it should definitely rename the file.

docs/00-getting-started.md
```diff
@@ -53,15 +59,17 @@ Once you have one, simply log in and go to the [Azure Portal](https://portal.azu
...
 copy secrets.yaml.template secrets.yaml
 ```
 
 Linux and Mac OS
 ```sh
+cd .thunderbolt
 cp secrets.yaml.template secrets.yaml
```
jiata: cp? it should just give you secrets.yaml and not secrets.yaml.template

paselem: agreed. If we're going down the 'init' path then we should just create the secrets.yaml file. The template file will be distributed with our package (I suppose) and copied into the .thunderbolt on the init call.

jafreck: yeah, the latest commit fixed this already. Now `azb spark init` automatically renames the secrets file for you. These docs should be updated to reflect that.

dtde/spark/cli/init.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+def setup_parser(parser: argparse.ArgumentParser):
+    pass
+
+def execute(args: typing.NamedTuple):
+    config_src_path = constants.INIT_DIRECTORY_SOURCE
+    config_dest_path = constants.INIT_DIRECTORY_DEST
+
+    copy_tree(config_src_path, config_dest_path, update=1)
```
paselem: What happens if the source path already exists?

jafreck: Might be misunderstanding the question, but config_src_path exists when the repo is pulled. It is the config/ directory that is included in the repo.

config_dest_path is the .thunderbolt directory we are creating in the cwd. If that directory already exists and contains files that are in config_src_path, the files are left untouched. So running `azb spark init` twice in the same directory will not overwrite existing files.

PR #80, opened 2017-09-18
Bug: Updated confusing cluster create success message

### LINKED ISSUE #78: Change print text after cluster create command so that it doesn't say "cluster created successfully"
Gotten feedback that this is a confusing message - conceptually it is correct (cluster of 0 nodes is created), but it is confusing. Instead it should say "You cluster is being created" something like that.

jafreck:
Fix #78


jiata:
lgtm

dtde/spark/cli/cluster_create.py
```diff
@@ -67,6 +67,9 @@ def execute(args: typing.NamedTuple):
         args.docker_repo,
         args.wait)
 
-    log.info("Cluster created successfully.")
+    if args.wait:
+        log.info("Cluster created successfully.")
+    else:
+        log.info("Cluster is being created.")
```
paselem: Minor: as per our conversation, you can add the cluster id to the log line, but may not be required since it will printed by line 44.

paselem: Hrm... actually, is there a bug here if the create_cluster fails? Not sure if we throw an exception, but if the cluster is not created for some reason (say a conflicting cluster name) won't line 71 report it was successful incorrectly?

jafreck: I don't believe so. The previous line makes an API call, and if anything goes wrong with that API call, an error is thrown. So these lines won't execute in the case of a bad configuration.

PR #83, opened 2017-09-19
update log to be more representative of state

paselem:


jafreck:
This is fixing the same issue as #80. Perhaps we should combine the solutions?

paselem:
Closing as #80 also addresses the same issue.

PR #85, opened 2017-09-20
Bug: Clean up error messages

jafreck:
Fix: #84


dtde/error.py
```diff
@@ -1,6 +1,19 @@
...
     pass
+
+
+class ThunderboltErrorException(Exception):
+    def __init__(self, message: str = None):
+        super().__init__()
+        self.message = message
+
+
+class AzureApiInitError(Exception):
```
timotheeguerin: can you make azure api just inherit from thunderbolt error exception

PR #93, opened 2017-09-22
Feature: Support for multiple custom scripts and master only scrips

### LINKED ISSUE #89: Only one custom script is allowed, add support for multiple

### LINKED ISSUE #91: Allow custom scripts to execute on the master only

jafreck:
Fix #91 
Fix #89


custom-scripts/simple.sh
```diff
@@ -1,3 +1,8 @@
 #!/bin/bash
 
-echo "This is a custom script!"
+if  [ "$IS_MASTER" = "1" ]; then
+    echo "This is a custom script running on just the master!"
+fi
+
+echo "This is a custom script running all workers!"
```
timotheeguerin: all workers and the master

dtde/clusterlib.py
```diff
@@ -80,24 +80,25 @@ def docker_run_cmd(docker_repo: str = None) -> str:
 def generate_cluster_start_task(
         cluster_id: str,
         zip_resource_file: batch_models.ResourceFile,
-        custom_script: str = None,
+        custom_scripts: list = None,
```
timotheeguerin: List[str] I think is better

custom-scripts/simple.sh
jiata: typo: running on*

jiata:
this is not really what i had in mind from a design perspective

Instead of using the "IS_MASTER" environment variable and make users write logic in bash (ex. if Master Node), i was thinking more along the lines of having the user specify which is a master-node-only script and which is a every-node script.

example:
azb spark cluster create --id my_cluster \\
    **--master-script custom-script-1.sh** \\
    **--worker-script custom-script-2.sh**

(maybe "worker-script" isn't the best name)

Intuitively i think this is a better experience as i doesn't require users to know about the environment variable "IS_MASTER" and write logic around that in bash.

@paselem  what do you think

jiata:
also, is #60 part of this feature? For many cases, custom script on master is not useful without the ability to open custom ports.

timotheeguerin:
I think `IS_MASTER` give the user more flexibility to run a custom script however they might want.

If there is something that needs to setup on both worker and master with an extra setup on a master this is easily done this way.

jiata:
Yeah, i like that idea.

but what are the scenarios you had in mind where you would need that flexibility of IS_MASTER?

Separating makes sense to me just in terms of the scenarios that i can think of:
- ADLS, WASB, S3, custom package installs (run on all nodes)
- livy, jupyter, spark magic, r-studio-server (only on master node)

paselem:
@jiata - I was thinking more along the lines of what you specified except just _always_ point to a directory. If there are files there, we would automatically run them. That is a pretty common practice for lots of over linux-like tooling and OS settings. We could have two directories, one for master scripts, one for non-master scripts. This would alleviate users from having to specify the if condition in the script which is really an 'us' problem.

Also, another issue that might come up is whether or not ordering matters. I'm not sure how someone would specify that and how we would guarantee it. Possibly register a collection of scripts in the config file or (as per Jacob's suggestion) just write an uber script that calls the other scripts in the order it wants.

jiata:
I think we should also add the ability to open custom ports (on cluster create) as part of this PR. 
```
custom_scripts:
- script: livy_script.sh
  is_master: true
  ports: [8998]
- script: jupyter_script.sh
  is_master: true
  ports: [8888]
```
   
Something like this ^

docs/11-custom-scripts.md
```diff
@@ -3,10 +3,41 @@ Custom scripts allow for additional cluster setup steps when the cluster is bein
 when you want to modify the default cluster configuration for things such as modifying spark.conf, adding jars
 or downloading any files you need in the cluster.
 
+You can specify the location of custom scripts on your local machine in `.thundebolt/cluster.yaml`. If you do not have a `.thunderbolt/` directory in you current working directory, run `azb spark init` or see [Getting Started](./00-getting-started). Note that the path can be absolute or relative to your current working directory.
+
+The custom scripts can be configured to run on the Spark master only, the Spark workers only, or all nodes in the cluster (Please note that by default, the Spark master node is also a Spark worker). For example, the following custom-script configuration will run 3 custom scripts:
+
```
jiata: maybe worth pointing out here that the scripts will be executed in the order that the user defines it in the cluster.yaml

jafreck:
The command line experience is broken as of the latest commit (this still needs to be fixed). I think it should probably be removed entirely and custom scripts should only be supported through cluster.yaml.

docs/11-custom-scripts.md
jafreck: I do that in the following paragraph, but it might be more clear to put that info here.

jiata: yeah i noticed - what you have below is useful i think. But i think it could be stated more explicitly, like "scripts will execute in the order that they are listed".

config/cluster.yaml
```diff
@@ -18,7 +18,12 @@ username: spark
 # docker_repo: <name of docker image repo (for more information, see https://github.com/Azure/thunderbolt/blob/master/docs/12-docker-image.md)>
 docker_repo: jiata/thunderbolt:0.1.0-spark2.2.0-python3.5.4
 
-# custom_script: <path to custom script to run on each node> (optional)
+# # optional custom scripts to run on the Spark master, Spark worker or all nodes in the cluster
+# custom_script: 
+#   - script: </path/to/script.sh or /path/to/script/directory/>
+#     location: <master/worker/all-nodes>
```
paselem: rename 'location' to 'runOn'

custom-scripts/simple.sh
```diff
@@ -1,3 +1,8 @@
 #!/bin/bash
 
-echo "This is a custom script!"
+if  [ "$IS_MASTER" = "1" ]; then
+    echo "This is a custom script running on just the master!"
+fi
+
+echo "This is a custom script running all workers and the master!"
```
paselem: Maybe add that we currently always deploy a worker on the master node, so we do not support a 'worker only' install.

paselem: Nm... just saw docs below.

docs/11-custom-scripts.md
```diff
@@ -3,10 +3,41 @@ Custom scripts allow for additional cluster setup steps when the cluster is bein
...
+Directories may also be provided in the custom-scripts section of `.thunderbolt/cluster.yaml`. 
+
+```yaml
+custom-scripts:
+    - script: /custom-scripts/
+      location: all-nodes
+```
+
+The above configuration takes the absolute path `/custom-scripts/` and upload every file within it. These files will all be executed, although order of exection is not guarenteed. If your custom scripts have dependencies, specify the order by providing the full path to the file as seen in the first example.
```
paselem: typo "upload" -> "uploads"

dtde/clusterlib.py
```diff
@@ -138,9 +135,44 @@ def generate_cluster_start_task(
...
+def upload_custom_script_config(custom_scripts=None):
+    with open(os.path.join(constants.CUSTOM_SCRIPTS_DEST, 'custom-scripts.yaml'), 'w+') as f:
+        f.write(yaml.dump(custom_scripts, default_flow_style = False))
+
+
+def move_custom_scripts(custom_scripts=None):
+    if custom_scripts is None:
+        return
+    
+    os.mkdir(os.path.join(constants.ROOT_PATH, 'node_scripts', 'custom-scripts'))
```
paselem: If the directory exists, do we need to clean it up before copying files over? Can files from a previous cluster create still linger?

paselem: ... Or is that what line 203 does? May be worth doing it here as a first step in case the process exists unexpectedly between lines 197 and 203 since it shouldn't harm anything and provides an extra guarantee.

dtde/config.py
```diff
@@ -139,8 +139,8 @@ def _merge_dict(self, config):
         if 'password' in config and config['password'] is not None:
             self.password = config['password']
 
-        if 'custom_script' in config and config['custom_script'] is not None:
-            self.custom_script = config['custom_script']
+        if 'custom_scripts' in config and config['custom_scripts'] not in [[None], None]:
```
paselem: 'custom_scripts' used several times. Should be a constant.

node_scripts/docker_main.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 # This file is the entry point of the docker container.
-# It will run the custom scripts if present and start spark.
+# It will setup WASB and start Spark.
```
paselem: worth commenting that it currently uses the same storage account as the one configured in the secrets.yaml (or conversely that the one in secrets.yaml is used for moving data around for this tool to work)

node_scripts/install/scripts.py
```diff
@@ -0,0 +1,73 @@
...
+def _run_scripts_dir(root: str = None):
+    try:
+        for path, subdirs, files in os.walk(root):
+            for name in files:
+                script_path = os.path.join(path, name)
+                _run_script(script_path)      
+    except FileNotFoundError as e:
+        print(e)
+    except IOError as e:
+        print(e)
```
paselem: final except catch all?

dtde/clusterlib.py
jafreck: I think we should actually clean up before we copy, and after. This will remove lingering files from a previous create, copy over the expected files, upload them and then delete them.

The only way someone could have lingering files, though, is if the program crashed between 197 and 203, since 203 does remove all the files.

dtde/config.py
jafreck: I don't understand what this means.

node_scripts/install/scripts.py
jafreck: This error checking is just meant to ensure that the directory of scripts exists. However, we probably want to catch errors in _run_script() though so user script crashes are caught.

PR #97, opened 2017-09-22
Bug: removed printed errors when spark conf files don't exist

### LINKED ISSUE #87: Suppress error when spark-defaults and spark-env are not found

jafreck:
Fix #87


dtde/config.py
```diff
@@ -19,15 +19,15 @@ def load_spark_config():
                 os.path.join(constants.DEFAULT_SPARK_CONF_DEST, 'spark-defaults.conf'))
         log.info("Loaded spark-defaults.conf")
     except Exception as e:
-        print(e)
+        pass
```
paselem: Fine for now, but going forward this kind of stuff should go to a log file.

PR #108, opened 2017-09-27
Bug: fixed --no-wait not being respected

### LINKED ISSUE #98: [azb spark cluster create --id my_cluster --no-wait] unexpected behavior

jafreck:
Fix #98


jiata:
So if i have my username declared in the cluster.yaml file, and i use the `--no-wait` flag, it will
1. try to not wait, 
2. and then the catch that it is and error (on this line: https://github.com/Azure/thunderbolt/blob/master/dtde/config.py#L186)

is that right?

jafreck:
yes that is right.

PR #109, opened 2017-09-27
Bug: prompt for confirmation of cluster id on delete

### LINKED ISSUE #90: Cluster delete should prompt for confirmation

jafreck:
Fix #90


jafreck:
Are we okay with uncolored text for the prompt?

jiata:
would it make more sense for it to be a y/n question instead of re-typing the cluster id? not really sure, but just what i would expect intuitively.

jafreck:
Typing y/n is a bit more intuitive. I decided to require typing the full cluster id as that is what the Azure portal requires when you delete a cluster.

Either is fine imo

dtde/spark/cli/cluster_delete.py
```diff
@@ -12,7 +12,14 @@ def setup_parser(parser: argparse.ArgumentParser):
 
 def execute(args: typing.NamedTuple):
     cluster_id = args.cluster_id
+
+    confirmation_cluster_id = input("Please confirm the id of the cluster you wish to delete: ")
```
paselem: As part of this change, we should have a --force flag to override this option as it may get annoying for some users (like me :))

jiata: yeah. i think this is actually better than y/n

paselem: I think we should have both. If we remove the y/n then we should not need a force flag.

jafreck: @jiata Are you saying we have users confirm by typing the cluster id (instead of y/n), and also have a --force flag?

PR #112, opened 2017-09-27
Bug: change default cluster create experience to require --id cli parameter

### LINKED ISSUE #86: Remove cluster_id from default cluster configuration
Recommended workflow is changing so that cluster id should be passed on the command line, not set in the cluster.yaml configuration file.

### LINKED ISSUE #117: IDs should be declared as a param (and not in cluster.yaml)
I remember we chatted about enforcing that people set their cluster id as a param, and not a config

https://github.com/Azure/thunderbolt/blob/master/dtde/config.py#L176
https://github.com/Azure/thunderbolt/blob/master/config/cluster.yaml#L4

not sure if there are other places that this needs to be updated.

jafreck:
Fix #86 
Fix #117


jiata:
still need to update here: https://github.com/Azure/thunderbolt/blob/master/dtde/config.py#L176

PR #115, opened 2017-09-27
Bug: add spark conf defaults

### LINKED ISSUE #111: Add default files for spark-env and spark-defaults

jafreck:
Fix #111 

Added defaults for spark-env.sh and spark-defaults.conf. These are pulled directly from the spark container.

Should we annotate the file at all? Possibly note which of the settings won't work?


config/spark-defaults.conf
```diff
@@ -0,0 +1,28 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+# Default system properties included when running spark-submit.
+# This is useful for setting default environmental settings.
+
+# Example:
+# spark.master                     spark://master:7077
```
jiata: we may want to curate this out of the config or make a comment stating that this is not applicable.

config/spark-env.sh
```diff
@@ -0,0 +1,62 @@
...
+# - SPARK_LOCAL_IP, to set the IP address Spark binds to on this node
+# - SPARK_PUBLIC_DNS, to set the public dns name of the driver program
+
+# Options read by executors and drivers running inside the cluster
+# - SPARK_LOCAL_IP, to set the IP address Spark binds to on this node
+# - SPARK_PUBLIC_DNS, to set the public DNS name of the driver program
+# - SPARK_LOCAL_DIRS, storage directories to use on this node for shuffle and RDD data
+# - MESOS_NATIVE_JAVA_LIBRARY, to point to your libmesos.so if you use Mesos
+
+# Options read in YARN client mode
```
jiata: not applicable

config/spark-env.sh
```diff
@@ -0,0 +1,62 @@
...
+# - HADOOP_CONF_DIR, to point Spark towards Hadoop configuration files
+# - SPARK_EXECUTOR_CORES, Number of cores for the executors (Default: 1).
+# - SPARK_EXECUTOR_MEMORY, Memory per Executor (e.g. 1000M, 2G) (Default: 1G)
+# - SPARK_DRIVER_MEMORY, Memory for Driver (e.g. 1000M, 2G) (Default: 1G)
+
+# Options for the daemons used in the standalone deploy mode
+# - SPARK_MASTER_HOST, to bind the master to a different IP address or hostname
```
jiata: not applicable

jafreck:
Adding comments in the file to alert users to fields that don't apply seems fine, but we should also probably compile a running list of Spark configuration settings that we don't support somewhere in the docs. That way if they just replace this with their own files, they still have somewhere to look to see what is/isn't supported.

config/spark-defaults.conf
```diff
@@ -0,0 +1,28 @@
...
+# spark.eventLog.enabled           true
+# spark.eventLog.dir               hdfs://namenode:8021/directory
+# spark.serializer                 org.apache.spark.serializer.KryoSerializer
+# spark.driver.memory              5g
+# spark.executor.extraJavaOptions  -XX:+PrintGCDetails -Dkey=value -Dnumbers="one two three"
+spark.jars                 /home/spark-current/jars/azure-storage-2.0.0.jar,/home/spark-current/jars/hadoop-azure-2.7.3.jar
```
paselem: Add a comment here that we pre-load wasb jars.

Going forward we'll (probably) also pre-load adls jars as well. I *think* s3 and s3a should be there too, but if it gets to be too much, we should probably have samples or plugins for it.

PR #116, opened 2017-09-28
Feature: prompt for password

### LINKED ISSUE #99: sshkey not working from config when using the ssh cmd
When I run `azb spark cluster create --id test` by default it should create the cluster, wait, and then create the user for the cluster (with the specified SSH_key)

However, this is broken today. After waiting for the cluster to spin up, i should be able to run `azb spark cluster ssh --id test`, and it should just use the ssh key. Instead, i get prompted with inputing a password.

This makes me think that creating the cluster with the user did not work properly.

### LINKED ISSUE #110: prompt for password if no ssh-key is provided for cluster create and cluster ssh

jafreck:
Fix #110 
Fix #99


config/secrets.yaml.template
```diff
@@ -11,6 +11,8 @@ storage:
     storageaccountsuffix: core.windows.net
 
 default:
-    # SSH public key used to create a user and connect to a server.
-    # It can either be the public key itself(ssh-rsa ...) or the path to the ssh key.
+    # SSH keys used to create a user and connect to a server.
+    # The public key can either be the public key itself(ssh-rsa ...) or the path to the ssh key.
+    # The private key must be the path to the key.
```
jiata: do we require them to enter a private ssh key even if its called *id_pub*?

jafreck: no, ssh_priv_key is entirely optional. If it is not specified, the ssh client they are using will still look for a private key using the standard paths & file names it expects (which includes id_pub in the ~/.ssh directory.

paselem: We should document that it is optional and perhaps even remove it from the template and only have that in documentation.

jafreck: yeah -- I think it is a real edge case that it would be used.

config/secrets.yaml.template
```diff
@@ -11,6 +11,7 @@ storage:
     storageaccountsuffix: core.windows.net
 
 default:
-    # SSH public key used to create a user and connect to a server.
-    # It can either be the public key itself(ssh-rsa ...) or the path to the ssh key.
+    # SSH keys used to create a user and connect to a server.
+    # The public key can either be the public key itself(ssh-rsa ...) or the path to the ssh key.
```
paselem: minor: space missing here "itself(ssh-rsa ...)"

dtde/clusterlib.py
```diff
@@ -198,6 +199,17 @@ def create_cluster(
...
+    # Get user ssh key, prompt for password if necessary
+    ssh_key = ssh.get_user_public_key(ssh_key)
+    if username is not None and password is None and ssh_key is None:
+        password = getpass.getpass("Please input a password for user {0}: ".format(username))
+        confirm_password = getpass.getpass("Please confirm your password for user {0}: ".format(username))
+        if password != confirm_password:
+            raise ThunderboltError("Password confirmation did not match, please try again.")
+        if not password:
+            raise ThunderboltError(
+                "Cannot add user to cluster. Need to provide a ssh public key in .thunderbolt/secrets.yaml or password.")
```
paselem: Can it also be done via a command line argument?

jafreck: Yeah, you can do it via command line as well. This error is actually a bit odd. This would only happen if the user didn't supply a password or ssh key for the user they are trying to create, and then when prompted to input a password, just hit enter twice (didn't input a password). So I'm not sure that this is even the best description of the issue. I'll update it to be more reflective.

PR #121, opened 2017-09-28
Bug: clearer error message for bad vm size

### LINKED ISSUE #23: Handle error when parameter --vm-size is invalid
Current error is cryptic and not useful

jafreck:
Fix #23


dtde/cli.py
```diff
@@ -31,6 +31,10 @@ def main():
     try:
         run_software(args)
     except batch_error.BatchErrorException as e:
+        try:
```
paselem: You should use util.print_batch_exception here isntead.

PR #133, opened 2017-10-02
Bug: simplify secrets loading

### LINKED ISSUE #120: Change implementation of SecretsConfig so it only needs to be called once
Currently, the SecretsConfig object is being created multiple times which could lead to unexpected behavior if the file changes during execution.

Potential Solution:
- Make a ClusterLib object and move functions in clusterlib.py under it
- read the secrets file once on the object's initialization
- pass the necessary information read from the secrets.yaml config file by parameter to functions
- each command line endpoint should instantiate this object
- Do the same for joblib.py

Remaining issues:
- what should we do with azure_api.py? what is the separation in responsibilities between this and ClusterLib?

jafreck:
Fix #120


aztk/aztklib.py
```diff
@@ -0,0 +1,37 @@
+from aztk.clusterlib import ClusterLib
+from aztk.joblib import JobLib
+import aztk.azure_api as azure_api
+import aztk.config as config
+import aztk.error as error
+
+
+class AztkLib:
```
timotheeguerin: Hhm I think this Lib suffix doesn't feel right. Maybe just `Aztk` or something more verbose(not sure about this though) 
@paselem @jiata what do you think?

aztk/aztklib.py
```diff
@@ -0,0 +1,37 @@
...
+        batch_config = azure_api.BatchConfig(
+            account_key = secrets_config.batch_account_key,
+            account_name = secrets_config.batch_account_name,
+            account_url = secrets_config.batch_service_url
+        )
+
+        self.batch_client = azure_api.make_batch_client(batch_config)
+        self.blob_client = azure_api.make_blob_client(blob_config)
+
+        self.clusterlib = ClusterLib(
```
timotheeguerin: would just do self.cluster here same below for joblib

aztk/clusterlib.py
```diff
@@ -10,524 +10,535 @@
...
+
+    def __init__(self, batch_client, blob_client, batch_config, blob_config, secrets_config):
+        self.batch_client = batch_client
+        self.blob_client = blob_client
+        self.batch_config = batch_config
+        self.blob_config = blob_config
+        self.secrets_config = secrets_config
+
+
+    def cluster_install_cmd( self,
```
timotheeguerin: space before self

aztk/aztklib.py
jiata: i would just use `Aztk` if that's possible

jafreck: So we are landing on `aztk.cluster.cluster_function()` and `aztk.job.job_function()`?

PR #138, opened 2017-10-04
Feature: add-user accept ssh-key path and prompt for password

### LINKED ISSUE #132: Add-User only takes ssh-key, not path to ssh-key
To add a user with an ssh-key, you must pass the full base-64 encoded public key. You should be able to pass a path to the key as is the case in `azb spark cluster create`.

jafreck:
Fix #132


aztk/clusterlib.py
```diff
@@ -359,7 +366,7 @@ def pretty_dedicated_node_count(self, cluster: ClusterModel)-> str:
             return '{}'.format(cluster.dedicated_nodes)
 
     def pretty_low_pri_node_count(self, cluster: ClusterModel)-> str:
-        if (cluster.pool.allocation_state is batch_models.AllocationState.resizing\
```
jiata: was `\` required?

jafreck: No -- not required, just a formatting ting. Right above this, pretty_dedicated_node_count() does the same thing without a backslash.

PR #142, opened 2017-10-04
Feature: ssh directly into container

### LINKED ISSUE #140: ssh directly into docker image (as opposed to host)

jafreck:
Fix #140


aztk/config.py
```diff
@@ -273,10 +274,13 @@ def _merge_dict(self, config):
         if 'jupyter_port' in config and config['jupyter_port'] is not None:
             self.jupyter_port = config['jupyter_port']
 
-        if 'connect' in config and config['connect'] is False:
-            self.connect = False
+        if 'host' in config and config['host'] is not None:
```
timotheeguerin: you can do `config.get('host') is not None` instead of this double check here

aztk/clusterlib.py
```diff
@@ -359,7 +366,7 @@ def pretty_dedicated_node_count(self, cluster: ClusterModel)-> str:
             return '{}'.format(cluster.dedicated_nodes)
 
     def pretty_low_pri_node_count(self, cluster: ClusterModel)-> str:
-        if (cluster.pool.allocation_state is batch_models.AllocationState.resizing\
+        if (cluster.pool.allocation_state is batch_models.AllocationState.resizing
```
timotheeguerin: removed \ only on this line?

jafreck: I don't think it's necessary. It is done the same way in pretty_dedicated_node_count() just above this.

timotheeguerin: i mean there is one next line too that is not needed

PR #145, opened 2017-10-04
Bug: fix same cluster id error

### LINKED ISSUE #144: Creating a cluster with the same id twice gives error about jobs
When a user tries to create a cluster with the same id as an existing cluster, the following error occurs:

```sh
Exception encountered:
The specified job already exists.
RequestId:2f1bdde5-40d3-4768-ab05-442a7c0dccd
Time:2017-10-04T21:48:27.4733404Z
````
The error should be about pool id, not jobs.

jafreck:
Fix #144 

New error:
"A cluster with the same id already exists. Use a different id or delete the existing cluster by running `aztk spark cluster delete --id <id>`"


timotheeguerin:
I think we should catch that error down and do something like `raiseAztkError("A cluster with this id already exist.")` because here we just change the message to be pool instead of job(which is better but still not great)

PR #149, opened 2017-10-05
Feature: global flag for init

### LINKED ISSUE #143: Add global flag and configuration to aztk spark init
- `aztk spark init` should have a --global flag that creates a `.aztk/` directory in the user's home directory.
- Order of precedence for configuration settings is:
    1. command line arguments
    2. `cwd/.aztk/`
    3. global `~/.aztk/`

jafreck:
Fix #143


aztk/config.py
```diff
@@ -76,37 +75,17 @@ def load_secrets_config(self, path: str=constants.DEFAULT_SECRETS_PATH):
 
     def _merge_dict(self, secrets_config):
         # Ensure all necessary fields are provided
-        try:
-            self.batch_account_name = secrets_config['batch']['batchaccountname']
-        except KeyError:
-            raise error.AztkError(
-                "Please specify a batch account name in your .aztk/secrets.yaml file")
```
timotheeguerin: are you validating anywhere now if the config is not valid?

aztk/config.py
```diff
@@ -115,12 +94,16 @@ def _merge_dict(self, secrets_config):
...
         default_config = secrets_config.get('default')
-
         # Check for ssh keys if they are provided
         if default_config:
             self.ssh_priv_key = default_config.get('ssh_priv_key')
             self.ssh_pub_key = default_config.get('ssh_pub_key')
 
+    def load(self):
+        # read global ~/secrets.yaml
+        self._load_secrets_config(os.path.join(constants.HOME_DIRECTORY_PATH, '.aztk', 'ssh.yaml'))
```
timotheeguerin: is this required? would it fail if there is no .aztk in my home dir?

aztk/spark/cli/init.py
```diff
@@ -6,19 +6,26 @@
 
 
 def setup_parser(parser: argparse.ArgumentParser):
-    pass
+    parser.add_argument('--global', dest='global_flag', action='store_true')
```
timotheeguerin: add a help text here maybe

aztk/config.py
jafreck: Yeah, validation is done in azure_api.py still. This was duplicating that so I removed it.

aztk/config.py
jafreck: No, it won't fail. If the path provided does not exist, _load_secrets_config just returns. 

That is done here: https://github.com/jafreck/aztk/blob/feature/global-config/aztk/config.py#L64

PR #164, opened 2017-10-11
Bug: remove unnecessary error

### LINKED ISSUE #163: remove unnecessary error in node stdout.txt
If no custom scripts are given, an error is logged on the node when [this line](https://github.com/Azure/aztk/blob/master/node_scripts/setup_node.sh#L42) executes.

Since the files are chmod'd in the container before they are run this line is not necessary at all.

jafreck:
Fix #163


node_scripts/setup_node.sh
```diff
@@ -39,7 +39,6 @@ fi
 # Unzip resource files and set permissions
 apt-get -y install unzip
 chmod 777 $AZ_BATCH_TASK_WORKING_DIR/docker_main.sh
-chmod -R 777 $AZ_BATCH_TASK_WORKING_DIR/custom-scripts
```
timotheeguerin: wait isn't that needed to run the custom scripts?

jafreck: No - we chmod each script on the node before we run it, (here)[https://github.com/jafreck/aztk/blob/bug/remove-chmod-error/node_scripts/install/scripts.py#L47]. This was just causing an unnecessary error if the custom-scripts directory didn't exist.

PR #166, opened 2017-10-13
Bug: validate the keys provided in configuration files

### LINKED ISSUE #162: Invalid keys in yaml files should result in an error instead of passive failure

jafreck:
Fix #162


jiata:
why do we call these 'keys'? aren't they parameters. I guess i think of secrets when we say 'keys'

jiata:
whoops, accidentally closed

jafreck:
These are the keys of the dictionary we are merging.

PR #168, opened 2017-10-13
Fix cluster yaml typo 

brnleehng:
Users who uncomment the cluster yaml will have issues with custom scripts. 

Probably need to validate the keys for cluster yaml template


jafreck:
Key validation for cluster.yaml and ssh.yaml is in #166

PR #182, opened 2017-10-19
Bug: fix error with pyenv export on container

### LINKED ISSUE #181: bash export error appears when sshing into container
```sh
bash: export: `export PATH="//.pyenv/shims:${PATH}"
export PYENV_SHELL=bash
source '//.pyenv/libexec/../completions/pyenv.bash'
command pyenv rehash 2>/dev/null
pyenv() {
  local command
  command="$1"
  if [ "$#" -gt 0 ]; then
    shift
  fi

  case "$command" in
  rehash|shell)
    eval "$(pyenv "sh-$command" "$@")";;
  *)
    command pyenv "$command" "$@";;
  esac
}': not a valid identifier
```

jafreck:
Fix #181


jafreck:
@jiata This will require updates to our images

jiata:
@jafreck  will do thx for this fix

PR #185, opened 2017-10-27
Feature: Flink

### LINKED ISSUE #186: Feature: Flink
Add support for Flink 

https://flink.apache.org/

jafreck:
added Flink support

Fix #186


aztk_sdk/utils/constants.py
```diff
@@ -0,0 +1,73 @@
...
+DEFAULT_FLINK_DOCKER_REPO = "jafreck/aztk:flink"
+DOCKER_FLINK_CONTAINER_NAME = "flink"
+
+# DOCKER
+DOCKER_SPARK_WEB_UI_PORT = 8080
+DOCKER_SPARK_WORKER_UI_PORT = 8081
+DOCKER_SPARK_JUPYTER_PORT = 8888
+DOCKER_SPARK_JOB_UI_PORT = 4040
+DOCKER_SPARK_HOME = "/home/spark-current"
+DOCKER_FLINK_HOME = "/opt/flink"
```
jiata: spark and flink should both be in `home` or `opt`. i think `opt` actually makes more sense / more common (which means that this fix is an item for me i guess)

docker-image/flink/Dockerfile
```diff
@@ -0,0 +1,102 @@
+# Ubuntu 16.04 (Xenial)
```
jiata: can we update this model off my new aztk spark vanilla dockerfile - in the branch: feature/new_containers

docker-image/flink/Dockerfile
```diff
@@ -0,0 +1,102 @@
...
+  wget -nv -O flink.tgz "$FLINK_TGZ_URL"; \
+  wget -nv -O flink.tgz.asc "$FLINK_ASC_URL"; \
+  \
+  export GNUPGHOME="$(mktemp -d)"; \
+  tar -xf flink.tgz --strip-components=1; \
+  rm flink.tgz; \
+  \
+  chown -R flink:flink .;
+
+# sshd
```
jiata: why do we need open ssh server

node_scripts/flink/docker_main.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
```
jiata: i feel like the entire node_scripts dir can be restructured a bit. A lot of the functions can be used for both spark and flink (like choose master, or wait for master, or setup ssh keys, etc...). The only main functions that need to be different are probably just __start_master__ and __start_worker__ and maybe some setup function to connect the cluster together right?

the directory is also structured in a confusing way. 

It is like this right now:
```
node_scripts
  /flink
    docker_main.sh
    /core
    /install
    ...
  docker_main.sh (for spark)
  /core (for spark)
  /install (for spark)
  ...
```

shouldn't it be more like this:
```
node_scripts
  /flink
  /spark
  /shared_stuff
  docker_main.sh
```

node_scripts/flink/docker_main.sh
```diff
@@ -0,0 +1,28 @@
...
+
+# This file is the entry point of the docker container.
+# It will install Flink
+# This script uses the storage account configured in .thunderbolt/secrets.yaml
```
jiata: .aztk/secrets.yaml

node_scripts/flink/docker_main.sh
jafreck: My plan was to move all the stuff in the root of node_scripts now into a `spark/` directory like you have in your second example.

I'm a little wary of having a `node_scripts/shared_stuff` though. Flink and Spark have incredibly similar setup processes, but that is more or less just a coincidence (or more likely since they are both Apache projects and Flink borrowed from some of Sparks design).

My idea around `node_scripts` is that each software's sub-directory will contain all and only the files needed to completely set up that software on the nodes. That way in the future if Flink diverges in it's setup processes from Spark, we only need to change the Flink `node_scripts`.

docker-image/flink/Dockerfile
jafreck: I took a look at that file, It has some Spark specific environment variables being set. I'm not sure what we should do because I feel like we should have a base Aztk image that all software build off of. So we would have `Ubuntu -> Aztk -> $Software Vanilla -> $Software Specific`, but I think to do that we have to go through the hoop of building base images.

PR #196, opened 2017-11-02
Bug: Fix docker auth for private repos

### LINKED ISSUE #188: Private Docker repo build hangs

jafreck:
Fix #188


jafreck:
the yaml parsing should take care of the case where the secrets.yaml file looks like this:

```yaml
docker:
    username:
    password:
```

The values are read as None (same as if they weren't there at all).

PR #200, opened 2017-11-02
only show clusters created with aztk in list

### LINKED ISSUE #199: List should filter on clusters created with AZTK

jafreck:
Fix #199


aztk_sdk/client.py
```diff
@@ -116,13 +116,20 @@ def __get_pool_details(self, cluster_id: str):
...
 
-    def __list_clusters(self):
+    def __list_clusters(self, software_metadata_key):
         """
             List all the cluster on your account.
         """
         pools = self.batch_client.pool.list()
 
-        return [pool for pool in pools]
+        aztk_pools = []
```
paselem: possibly cleaner way to structure your for loop with if conditions:

```python
important_airports = (airport for airport in airports if airport.is_important)
for airport in important_airports:
    # do stuff
```

source: https://stackoverflow.com/questions/2401785/single-line-for-loop-over-iterator-with-an-if-filter

jafreck: Are you telling me you *don't* like reading 5 deep nested for/if statements?

PR #201, opened 2017-11-02
Bug: fix pyenv init

jafreck:
Fix #182


aztk_sdk/spark/helpers/submit.py
```diff
@@ -33,6 +34,12 @@ def __app_submit_cmd(
 
     spark_home = constants.DOCKER_SPARK_HOME
 
+    # set file paths to correct path on container
+    files_path = '/batch/workitems/{0}/{1}/{2}/wd/'.format(cluster_id, "job-1", name)
```
paselem: minor: "job-1" should probably be a constant

jafreck: Can change that elsewhere, but this PR is meant to only be a fix to the Docker image so that pyenv is set up correctly. The branch should be updated to reflect that now.

PR #206, opened 2017-11-06
Feature: Spark GPU

### LINKED ISSUE #195: Host docker images on service account
Also add docker autobuild functionality

https://docs.docker.com/docker-cloud/builds/automated-build/#set-up-build-rules

### LINKED ISSUE #205: Add GPU support for Spark

jafreck:
Fix #205 
Fix #195


custom-scripts/numba.sh
```diff
@@ -0,0 +1,3 @@
+#! /bin/bash
+conda install -y cudatoolkit
+conda install -y numba
```
paselem: we might get this for free with the container we use for pyspark && sparklyR

examples/src/main/python/gpu/nubma_example.py
```diff
@@ -0,0 +1,32 @@
...
+from pyspark.sql import SparkSession
+
+@cuda.jit('(float32[:], float32[:])')
+def func(inp, out):
+    i = cuda.grid(1)
+    if i < out.size:
+        out[i] = inp[i] ** 2
+
+
+def gpu_work(xs):
```
paselem: would be cool if there was a CPU equivalent to compare with.

custom-scripts/numba.sh
jafreck: Looking at this: https://docs.anaconda.com/anaconda/packages/py3.5_linux-64, it seems numba does come by default but cudatoolkit does not.

Will update to reflect that.

examples/src/main/python/gpu/nubma_example.py
jafreck: updated to add a comparison

aztk_sdk/utils/helpers.py
```diff
@@ -16,6 +17,9 @@
 _STANDARD_ERROR_FILE_NAME = 'stderr.txt'
 
 
+def is_gpu_enabled(vm_size):
+    return bool(re.search('nv|nc', vm_size))
+
```
paselem: nitpick: two lines between methods?

custom-scripts/conda_cudatoolkit.sh
```diff
@@ -0,0 +1,2 @@
+#! /bin/bash
+conda install -y cudatoolkit
```
paselem: Do we always expect to install this via a custom script? Shouldn't this "just happen" if we detect a gpu sku?

examples/src/main/python/gpu/nubma_example.py
```diff
@@ -0,0 +1,60 @@
...
+
+start_time = time.time()
+cpu_some_trig(xs, ys)
+cpu_elapsed_time = time.time() - start_time
+
+start_time = time.time()
+cuda_some_trig(xs, ys)
+gpu_elapsed_time = time.time() - start_time
+
+print("GPU was", cpu_elapsed_time-gpu_elapsed_time, "seconds faster")
```
paselem: perhaps

```python
print("CPU time: {}".format(cpu_elapsed_time)) 
print("GPU time: {}".format(gpu_elapsed_time))
print("GPU performed {}% better than CPU".format(...))
```

examples/src/main/python/gpu/nubma_example.py
```diff
@@ -0,0 +1,60 @@
+import numpy as np
```
paselem: I really like this example. Can we do 2 things
1. Have a java/scala equivalent?
2. Have a jupyter notebook version as well? 

Probably also worth adding some code to try to detect if the GPU is present though before running this.

jafreck: [This should work for detecting GPUs with Numba](http://numba.pydata.org/numba-doc/0.19.1/cuda-reference/host.html#numba.cuda.is_available). It seems like most other gpu-enabled tools have similar functions so when I add java and scala examples, I'll make sure to include them.

docs/60-gpu.md
```diff
@@ -0,0 +1,8 @@
+# GPU
+
+Use GPUs to accelerate your Spark applications. By default, when you specify an Azure VM type that has a GPU, your Docker image will contain CUDA-8.0 and cuDNN-6.0. See [Docker Image](./12-docker-image.md) for more information about the AZTK Docker images.
```
paselem: Update "By default, when you specify an Azure VM type that has a GPU, your Docker image will contain CUDA-8.0 and cuDNN-6.0" 

to

"When using a [GPU enabled Azure VM](https://azure.microsoft.com/en-us/pricing/details/batch/), your docker image will contain CUDA-8.0 and cuDnn-6.0 by default."

docs/60-gpu.md
```diff
@@ -0,0 +1,8 @@
...
+
+AZTK uses Nvidia-Docker to expose the VM's GPU(s) inside the container. Nvidia drivers (ver 384) are installed at runtime.
+
+### Installation Location
+By default, CUDA is installed at `/usr/local/cuda-8.0`.
```
paselem: Can you add a sample or tutorial section?


# Tutorial

Create a cluster specifying a GPU enabled VM
```sh
aztk spark cluster create --id gpu-cluster --vm-size standard_nc6 --size 1
```

Submit your an application to the cluster that will take advantage of the GPU
```sh
aztk spark cluster submit --id gpu-cluster --name gpu-app ./examples/src/main/python/gpu/nubma_example.py
```

PR #213, opened 2017-11-09
add a default port for the history server

paselem:


aztk/config.py
```diff
@@ -218,6 +218,9 @@ def _merge_dict(self, config):
 
         if config.get('job_history_ui_port') is not None:
             self.job_history_ui_port = config['job_history_ui_port']
+        else:
+            # fall back to default port
+            self.job_history_ui_port = '18080'
```
paselem: I'm not sure if this should go here, or near line 184, and this option would simply override it. Thoughts?

jafreck: I think this belongs on line 184.

PR #215, opened 2017-11-10
Feature: HDFS plugin

jafreck:
#214


aztk_sdk/spark/utils/upload_node_scripts.py
```diff
@@ -102,6 +107,9 @@ def zip_scripts(blob_client, custom_scripts, spark_conf):
         zipf = __add_file_to_zip(zipf, spark_conf.spark_defaults_conf, 'conf', binary=False)
         zipf = __add_file_to_zip(zipf, spark_conf.spark_env_sh, 'conf', binary=False)
         zipf = __add_file_to_zip(zipf, spark_conf.core_site_xml, 'conf', binary=False)
+        # add ssh keys for hdfs
+        zipf = __add_str_to_zip(zipf, spark_conf.ssh_key_pair['pub_key'], 'id_rsa.pub')
```
paselem: should we fail the create cluster operation if we don't have these? (e.g. windows users who are not using a linux shell)

jafreck: We should always have these keys becuase they are generated on the fly [here](https://github.com/Azure/aztk/pull/215/files#diff-10d14a544bf3f6a6a9da910fb3ada2feR46).

PR #218, opened 2017-11-10
Bug: remove file loading from Aztk constructor

### LINKED ISSUE #217: Aztk constructor should not load file
https://github.com/Azure/aztk/blob/65a56dd65766abb41187605d42ecc23972cf943f/aztk/aztklib.py#L8

jafreck:
Fix #217


aztk/aztklib.py
```diff
@@ -4,9 +4,16 @@
 
 class Aztk:
     def __init__(self):
-        secrets_config = config.SecretsConfig()
-        secrets_config.load()
+        self.secrets_config = config.SecretsConfig()
+        self.client = None
 
+
+    def load(self):
```
paselem: Taking this a step further, you could argue that the method should take a path to the file you want to load as a parameter.

jafreck:
Handled by #231

PR #227, opened 2017-11-22
Bug: use merged username / password instead of args

bigonazzi:
Currently aztk cluster create merges username and passwords set in set in cluster.yaml, but it still require a password to be explicitly passed on the command line.


timotheeguerin:
are you saying that you are providing your password in the cluster.yaml but it still ask for it when running `cluster create`? This sounds like a bug however I don't think having your password in the cluster.yaml is a good idea to start with. If you want to save your credentials I recommend using the ssh public key.

bigonazzi:
@timotheeguerin yes, that's what's happening. I'm fully aware this isn't necessarily the best setup, but it's better than using args. I don't want to use a key, because it would just complicate our already complicated setup: store secrets in azure key-vault and run jobs from VSTS. The change is pretty straight-forward, could you assign a reviewer please?

bigonazzi:
@timotheeguerin Thanks for approving, I'm not authorised to push the change though. Will you take care of it?

paselem:
@jafreck - I will accept this merge, but we should think about add/moving this to the secrets.yaml file depending on how users want to share passwords. May be OK to leave it here as undocumented.

PR #231, opened 2017-11-23
Feature: Rename SDK

### LINKED ISSUE #191: Add docs for SDK

### LINKED ISSUE #217: Aztk constructor should not load file

### LINKED ISSUE #230: Rename aztk_sdk to sdk
Using the SDK should look like this:

```sh
import aztk.software
```
Not like this:
```sh
import aztk_sdk.software
```

jafreck:
Fix #230
Fix #217
Fix #191


examples/sdk/sdk_example.py
```diff
@@ -0,0 +1,111 @@
...
+    print(e.message)
+    sys.exit()
+
+# get details of specific cluster
+cluster = client.get_cluster(cluster_config.cluster_id)
+
+# # create a user for the cluster
+client.create_user(cluster.id, "sdk_example_user", "example_password")
+
+# create some apps to run
```
paselem: very minor - can we add one app which is scala/java as well?

setup.py
```diff
@@ -16,7 +16,7 @@
       ],
       entry_points=dict(
             console_scripts=[
-                  "{0} = aztk.cli:main".format(constants.CLI_EXE)
+                  "{0} = cli.cli:main".format(constants.CLI_EXE)
```
paselem: "cli.cli" looks a bit strange... shouldn't it just be cli:main ?

examples/sdk/sdk_example.py
jafreck: We don't distribute the compiled java/scala programs in the repo. So if I do a java/scala example, the script will either not work out of the box or the java/scala app will need to be commented out. 

Not sure which is more preferable, but I would lean towards leaving it commented out and having the script work by default.

cli/spark/aztklib.py
```diff
@@ -0,0 +1,23 @@
+import cli.config as config
+import aztk.spark
+
+
+def load_spark_client():
+    secrets_config = config.SecretsConfig()
```
timotheeguerin: just add a small doc here saying this method create the client for you using the local user environment

PR #240, opened 2017-11-29
Add support for VNETs

emlyn:
We need to run aztk clusters in a VNET in order to access other resources (e.g. a Hive metastore), so I made these modifications in order to make that work. Would you be interested in merging these changes upstream?

I added a `--subnet-id` command-line parameter and corresponding `cluster.yaml` key, that gets passed through in the network_configuration parameter when configuring the pool.

It turns out this is not supported by Azure Batch when using shared key authentication, so I had to add support for service principal authentication, both in aztk itself, and passed through to the node scripts so that they can also use the new credentials.

Let me know if you'd like anything done differently. I haven't updated the documentation yet, as I want to be sure the basic outline makes sense first to avoid having to rewrite it.


aztk_sdk/client.py
```diff
@@ -73,6 +77,10 @@ def __create_pool_and_job(self, cluster_conf, software_metadata_key: str, start_
             helpers.select_latest_verified_vm_image_with_node_agent_sku(
                 VmImageModel.publisher, VmImageModel.offer, VmImageModel.sku, self.batch_client)
 
+        network_conf = None
+        if cluster_conf.subnet_id is not None:
+            network_conf = batch_models.NetworkConfiguration(subnet_id=cluster_conf.subnet_id)
```
paselem: This should fail if the user is using Shared Key Auth

aztk/config.py
```diff
@@ -45,13 +50,20 @@ def _merge_dict(self, secrets_config):
...
             self.batch_service_url = batch.get('batchserviceurl')
+            self.batch_resource_url = batch.get('resourceurl')
 
         storage = secrets_config.get('storage')
         if storage:
             self.storage_account_name = storage.get('storageaccountname')
             self.storage_account_key = storage.get('storageaccountkey')
             self.storage_account_suffix = storage.get('storageaccountsuffix')
 
+        serviceprincipal = secrets_config.get('serviceprincipal')
```
paselem: Probably worth warning a user that one auth method will be used if more than 1 is added to this file. For example, if SharedKey && ServicePrincipal, then use ServicePrincipal.

aztk/config.py
```diff
@@ -11,11 +11,16 @@ def __init__(self):
         self.batch_account_name = None
         self.batch_account_key = None
         self.batch_service_url = None
+        self.batch_resource_url = None
```
paselem: Probably not worth having account name && resource url. Maybe not worth making the change yet, but we should only require resource url since it is more verbose. We can parse out the string if using SharedKey

aztk_sdk/utils/azure_api.py
```diff
@@ -37,16 +57,24 @@ def make_batch_client(batch_config: BatchConfig):
     """
     # Validate the given config
     _validate_batch_config(batch_config)
-    
-    # Set up SharedKeyCredentials
-    credentials = batch_auth.SharedKeyCredentials(
-        batch_config.account_name,
-        batch_config.account_key)
+
+    if batch_config.resource_url is None:
```
paselem: As per my comment above, maybe swap these around? Prefer ServicePrincipal over Shared Key?

aztk_sdk/client.py
emlyn: The API quickly returns quite a clear error without attempting to create the cluster in this case:
```
Server failed to authenticate the request. Make sure the value of Authorization header is formed correctly including the signature.
RequestId:f13dec02-df7b-4f95-a129-078bcd8ca2bf
Time:2017-11-30T09:40:09.7593552Z

AuthenticationErrorDetail:	The specified type of authentication SharedKey is not allowed when external resources of type Network are linked.
```
But happy to add an extra check if you still think it would be clearer.

aztk/config.py
emlyn: I agree that this does feel like duplication, but from the user's point of view, they only need to enter one or the other.
A benefit of having account name for shared key auth is that all the information necessary for configuring batch auth is together on one screen in the web UI, so requiring the resource URL would make set up more difficult for shared key users.

aztk/config.py
emlyn: The checks ([here](https://github.com/Azure/aztk/pull/240/files#diff-c64e530c75ccd0def3ecc50fbf949b2fR33)) currently use the presence of the resource url to decide whether to use shared key or service principal auth:
- If `resourceurl` is present, require that all `serviceprincipal` keys are present, and shared key account name and key must be absent
- if `resourceurl` is absent, ignore the `serviceprincipal` section (because in the future it may be used by other things) and require shared key details to be present

So it is not possible to specify both shared key and service principal auth for Batch (it will fail with `Both Batch resource url and account name are set in secrets.yaml config, can use either shared secret or service principal login, not both`.

I think this is clearer than allowing both to be specified, but then ignoring the shared key credentials.

Although if resource url and account name are merged into one key, we'd need to come up with something else anyway as this logic would no longer be possible.

emlyn:
Thanks for the feedback.
I wasn't sure about the logic for choosing between shared key and service principal auth, so any more feedback on that, in light of my replies above, would be appreciated. Whatever is chosen in the end would obviously need to be documented so that it is clear for the end user.

aztk_sdk/client.py
paselem: The main issue from my perspective is that a user may not know what a SharedKey is in the context of AZTK, unless we label it as such in the secrets file. The error introduces a techical concept of the service which is ideally not important to them. Perhaps as part of this change, we can change the configuration names to:

```yaml
 batch:
     batchserviceurl: 
     batchaccountname: 
     batchaccountsharedkey: 
```

and still support the old key names as well? I'm not sure if that makes it better because we still have a name mismatch between the portal and the yaml, but would at least help the user around naming. 

@jiata - do you have a strong opinion one way or the other on this?

aztk/config.py
paselem: This is a good point. I agree with you.

aztk_sdk/client.py
paselem: I had another thought entirely, which is to refactor the secrets file a little bit more to account for shared key or some form of AAD auth. One issue is that I think we're forgetting to consider the storage account as well which should also be managed by AAD if we're using Service Principal.

```yaml
sharedKey:
  batchaccounturl: my_batch_account_url
  batchaccountname: my_batch_account_name
  batchaccountkey: my_batch_account_key

  storageaccountname: my_storage_account_name
  storageaccountkey: my_stoarge_account_key
  storageaccountsuffix: my_storage_account_suffix

servicePrincipal:
  tenantid: my_tenant_id
  clientid: my_client_id
  credential: my_crenential

  batchaccountresourceid: my_batch_account_resource_id
  storageaccountresourceid: my_storage_account_resource_id
```

This gives us room to use other forms of authentication such as certificates or AAD applications, and makes it very explicit to the user which for of auth they are using. Further more, we could add comments into the sharedKey section mentioning that VNETS and MixMode clusters (dedicated && low pri in the same cluster) are not supported. I would expect that more advanced users will slowly graduate up from shared key to some form of AAD as they go from trying out the toolkit to using more advanced features.

emlyn: I think this makes sense, but there's just one issue: the batch account url is not specific to the shared key credentials, it is needed no matter what credentials are used (it is passed to the batch service client constructor along with the credentials ([here](https://github.com/Azure/aztk/pull/240/files#diff-c64e530c75ccd0def3ecc50fbf949b2fR77)) so I don't think it makes sense in the `sharedKey` section, but there isn't really another section for it now.
Also, if a change like this is made, should we continue to support the old keys for a while (possibly with a warning)?

emlyn: Also, I left out service principal auth for the storage account, as it was not necessary for configuring the VNET, and it was not immediately obvious to me how to do that. But if you've got some pointers to get started, I'm happy to make that change.
I noticed that the `BlockBlobService` can take a `sas_token`, and `ServicePrincipalCredentials.set_token` seems to get (?!) some sort of token, but not sure if these are the same thing.

jiata: @paselem I think the format you proposed - where we have a clear separation between Service principal and Shared key - makes the most sense to me, as it highlights that the auth methods have an either-or relationship. We should definitely add comments highlighting the limitations of shared keys too and encourage folks to use AAD

With the batch account url, i feel like we _shouldn't_ need it for AAD... I see your reference with passing it to the batch client constructor, but with AAD, are we not able to reconstruct that url? @paselem correct me if im wrong...

@emlyn re your comment on continuing support for Shared keys - My opinion is that we should continue to support it, while recommending AAD. Not supporting it would be a breaking change for existing users.

emlyn: It would be nice if it could reconstruct the batch account url, but I'm not sure where you would get all the information. It looks like the url is in the format `https://<accountname>.<region>.batch.azure.com`, so as well as the account name you would have to know the region somehow, and I'm not sure whether the last part is constant (e.g. might it be different in China?)

paselem: @emlyn - I took a quick look at the batch and storage constructors. It gets a bit confusing because both Batch and Storage have Service and Management SDKs. In both cases, the Management SDK is used to Manage accounts themselves (create, delete, list, get properties etc) while the Service SDKs are used to operate against objects in the service (e.g. GET/ADD/DELETE blobs for Storage, or List Pools for Batch).

What this means is that what we could do is use the management SDKs for both using the service principal and the fully qualified resource ID. We can then use that to pull the necessary information to create a Service SDK client constructor with the right parameters. A bit of pseudo code as an example:

```pythos
# Parse resource ID
subscriptionId, resourceGroup, accountName = 
  parseResourceId('/subscrtion/00000/resourcegroup/abc/storageAccount/myStorageAccount')

# Create a storage management Client
storageManagementClient = new StorageManagementSdkClient(credentials, subscriptionId, resourceGroup, accountName)

# Create a storage client
storageClient = new StorageSdkClient(accountName, storageManagementClient.getKeys())

# Party on
...

```

This would require a round-trip call for each storage call, but I suspect the performance penalty will not be very noticeable. We will obviously need to handle any authorization failures, but that needs to happen regardless with AAD auth.

Now, this change is becoming a bit more involved and we have some other big changes which are approved and about to go into master which are going to cause you a bit of a merging headache :). Here is my proposal, in order to unblock you lets take this change into a branch (say feature/VNET) and finish coding it up there. We will help you resolve any conflicts and address all of the multiple client layers and authentication logic. 

We are very eager to see this change make it into master, but I also want to make sure we are careful not to break backward compatibility with existing users, so giving it a branch to live in while we develop and test seems like the right path forward. I will create the branch now. Will you be able to create a new PR to point there instead?

emlyn: @paselem that sounds good, I'll create a new PR onto the branch. Thanks

paselem: @emlyn, @jafreck is going to create the branch soon. He will probably include all of your changes. We can continue to design / discuss from there. Thanks for all the contribution so far!

emlyn:
I've changed the base for this PR to be the feature/vnet branch, and merged in the changes from master

paselem:
@emlyn - thanks for doing that! I will close this PR for now and we can reopen a new one in the new branch.

emlyn:
@paselem what are the next steps to move this along, can I do anything?

paselem:
@emlyn - Sorry for the delay. I have started working on creating a Project to track this work. Please find it [here](https://github.com/Azure/aztk/projects/3). This is still a work in progress and some of the items have already been addressed in your change. The main things I think still require discussion is the formatting of the secrets.yaml file. I will submit my proposal in an issue and we can discuss what changes we want to make to it.

PR #241, opened 2017-11-29
Feature: Azure Files

paselem:
Allow users to mount file shares to their cluster for small data scenarios or scripts.


aztk_sdk/spark/helpers/create_cluster.py
```diff
@@ -69,14 +75,30 @@ def __get_docker_credentials(spark_client):
...
-    ret = [
+    shares = []
+
+    if file_mounts:
+        for mount in file_mounts:
+            # Create the directory on the node
+            shares.append('mkdir -p {0}'.format(mount.mount_path))
+
+            # Mount the file share
+            shares.append('mount -t cifs //{0}.file.core.windows.net/{2} {3} -o vers=3.0,username={0},password={1},dir_mode=0777,file_mode=0777,sec=ntlmssp'.format(
```
jafreck: Do we want to always mount with 777 permissions? Perhaps we should let the user specify this in the FileShare model.

paselem: that is a good point... perhaps they want read-only...

paselem: Well, after coding it up and testing it and finding out it doesn't work, I found this in the docs:

Is it possible to specify read-only or write-only permissions on folders within the share?
If you mount the file share by using SMB, you don't have folder-level control over permissions. However, if you create a shared access signature by using the REST API or client libraries, you can specify read-only or write-only permissions on folders within the share.

paselem: And a bit more research shows this: https://unix.stackexchange.com/questions/303811/can-super-user-write-into-read-only-files

Bascially root can do whatever they want, so having permissions on the file is pretty much useless since we run the container as root.

PR #252, opened 2017-12-07
Retry asking for password when it doesn't match or is empty

emlyn:
I recently created a cluster and made a typo when entering the password. The error message said "Password confirmation did not match, please try again." but then it didn't give me an opportunity to try again, and left the cluster in a state where I could not ssh in (I later realised I probably could have run `add-user` afterwards, but it wasn't obvious at the time).
I think it would make sense to re-prompt for the password in this case (or when the entered password is empty), what do you think?


jafreck:
Great idea! I think a limited number of attempts would be better though. Maybe we could do 3 retries like SSH allows. 

Also it would be nice to let users know that you can run `add-user` afterwards if they fail the 3 retries so there would be little confusion about how to recover from this bad state.

paselem:
+1 on limiting to 3 retries and letting users know about re-running the add-user command. Otherwise this looks good.

emlyn:
Thanks for the feedback. I agree that limiting the number of retries makes sense. How's this?

jafreck:
There is a bug here. If you give 3 mismatched passwords, the last password password will be used to create a user even though it did not match it's confirmation.

This is because the for else statement calls `log.error` on the failure case. But even then, the mismatched password is returned. 

Example:
Attempt 1:
1. password; 0
2. confirmation: 1

Attempt 2:
1. password; 1
2. confirmation: 2

Attempt 3:
1. password; 2
2. confirmation: 3

The user will be created with password '2', when the expected behavior would be a failure with no user created.

emlyn:
Oops, yes I've changed it to a throw.

emlyn:
I just rebased on the latest master.

paselem:
Thanks for the change @emlyn !

PR #257, opened 2017-12-08
Feature: fix numba example percentage

### LINKED ISSUE #255: Bug in numba example
1) file is titled: "nubma_example.py"
2) https://github.com/Azure/aztk/blob/master/examples/src/main/python/gpu/nubma_example.py#L64 this line doesn't calculate a percentage

jafreck:
Fix #255


examples/src/main/python/gpu/nubma_example.py
```diff
@@ -61,7 +61,7 @@ def cuda_some_trig(x, y):
 
     print("CPU time: {}".format(cpu_elapsed_time))
     print("GPU time: {}".format(gpu_elapsed_time))
-    print("GPU performed {}% faster than CPU".format((cpu_elapsed_time-gpu_elapsed_time)*100))
+    print("GPU performed {}% faster than CPU".format(((cpu_elapsed_time-gpu_elapsed_time)/gpu_elapsed_time)*100))
```
paselem: @jafreck - I think there are some other issues in this file. In this case, you're measuing the GPU performance of a single algorithm not using Spark. If I understand this correctly, what we should be measuing is the performance of
```py
rdd.mapPartitions(cpu_some_trig)
rdd.collect()
```
versus
```py
rdd.mapPartitions(gpu_some_trig)
rdd.collect()
```

As it is written now, I think we are simply doing calls directly against the CPU or GPU w/o using Spark as part of the benchmark.

jiata: I dont think we can just substitute the functions like that since cpu_some_trig and gpu_some_trig require two args, but yeah.. i believe mapPartitions() is the mapping function that will distribute some func across the partitions in the different nodes

PR #259, opened 2017-12-08
Feature: add default master filesystem HA

### LINKED ISSUE #258: Add Filesystem master HA
http://spark.apache.org/docs/2.0.1/spark-standalone.html#single-node-recovery-with-local-file-system

jafreck:
Fix #258

### DIFF
--- a/config/spark-env.sh
+++ b/config/spark-env.sh
@@ -44,3 +44,6 @@
 # - SPARK_IDENT_STRING  A string representing this instance of spark. (Default: $USER)
 # - SPARK_NICENESS      The scheduling priority for daemons. (Default: 0)
 # - SPARK_NO_DAEMONIZE  Run the proposed command in the foreground. It will not output a PID file.
+
+# DEFAULT FILESYSTEM HA
+export SPARK_DAEMON_JAVA_OPTS="-Dspark.deploy.recoveryMode=FILESYSTEM -Dspark.deploy.recoveryDirectory=/root/"


config/spark-env.sh
```diff
@@ -43,4 +43,7 @@
 # - SPARK_PID_DIR       Where the pid file is stored. (Default: /tmp)
 # - SPARK_IDENT_STRING  A string representing this instance of spark. (Default: $USER)
 # - SPARK_NICENESS      The scheduling priority for daemons. (Default: 0)
-# - SPARK_NO_DAEMONIZE  Run the proposed command in the foreground. It will not output a PID file.
\ No newline at end of file
+# - SPARK_NO_DAEMONIZE  Run the proposed command in the foreground. It will not output a PID file.
+
+# DEFAULT FILESYSTEM HA
+export SPARK_DAEMON_JAVA_OPTS="-Dspark.deploy.recoveryMode=FILESYSTEM -Dspark.deploy.recoveryDirectory=/root/"
```
paselem: Is there any advantages of using spark-env.sh vs. [spark-conf.defaults](https://spark.apache.org/docs/latest/spark-standalone.html#single-node-recovery-with-local-file-system) (or the other way around) for these kinds of options?

Also, complete ignorance on my part here, but does the SPARK_DAEMON_JAVA_OPTS apply to all of spark? The 'JAVA' part of the string has me a bit confused...

jafreck: Yeah, it applies to all. It has JAVA in it since Spark itself runs on the JVM and these options apply to the spark daemon.

Description of SPARK_DAEMON_JAVA_OPTS from the docs:
`JVM options for the Spark master and worker daemons themselves in the form "-Dx=y" (default: none).`

jafreck: Also as far as spark-env vs spark-defaults, the docs seem to suggest using spark-env:
`In order to enable this recovery mode, you can set SPARK_DAEMON_JAVA_OPTS in spark-env using this configuration`

I would imagine spark-defaults would work as well but I haven't tried it yet. Will test now.

paselem: As long as it works in all cases I'm OK with this change. Again, ignorance on my part, but since I'm not sure what most users will want, I would like to add most setting we default in a single place (like the jars we add for wasb). I don't have a strong opinion on this one, but if we can have all of our presets in a single place that _seems_ preferable to me. Either way I'll sign off as this seems to do the job.

jafreck: @paselem tested spark-defaults.conf and it works fine. I believe the only benefit to using spark-env.sh over spark-defaults.conf is that it is a shell script (so you can do more complicated variable setting). `spark-env.sh` does override `spark-defaults.conf`, though.

We should be fine putting all of our configuration pieces in spark-defaults.conf, at least so far. I definitely agree we shouldn't split up the configuration unless we need to.

PR #267, opened 2017-12-10
Feature: validate docker repo exists before cluster creation

### LINKED ISSUE #266: Validate the existence of the specified docker image before the cluster spins up
currently, if a bad docker-repo is configured, the cluster will still be provisioned. Failure will only occur once the cluster tries to download the image. At this point the cluster will just hang in a bad state.

A check for the existence of the docker repo should be made before the cluster is created, and an error should be thrown if the image is bad.

jafreck:
Fix #266

### DIFF
--- a/aztk/utils/helpers.py
+++ b/aztk/utils/helpers.py
@@ -3,6 +3,7 @@ import datetime
 import io
 import os
 import time
+import requests
 import re
 import azure.batch.batch_service_client as batch
 import azure.batch.batch_auth as batch_auth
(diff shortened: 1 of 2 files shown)


aztk/utils/helpers.py
```diff
@@ -26,6 +27,11 @@ def get_cluster(cluster_id, batch_client):
 
     return aztk.models.Cluster(pool, nodes)
 
+def validate_docker_repo(docker_repo):
+    repository, tag = docker_repo.split(":")
+    response = requests.get("https://index.docker.io/v1/repositories/{0}/tags/{1}".format(repository, tag))
```
paselem: This won't work for private docker repositories (e.g. Azure Container Repository - ACR).

The repository can be in the form:
private.repo.io/imagename:tag

We need to be able to handle any version of this. Probably worth looking into whether or not there is a docker API to check this.

PR #273, opened 2017-12-12
Feature: add feedback for cluster create wait

### LINKED ISSUE #261: on cluster created with `--wait`, there is no feedback from the cli, and it looks like it is hanging...
Currently, it looks like this until the cluster is created and *completely* ready.
screen shot 2017-12-09 at 5 04 57 pm

For now, if --wait is true, then we should just add a simple "please wait while your cluster is being provisioned" setting.

In the future, we should poll for pool status and provide progress feedback to the user.

jafreck:
Fix #261


cli/spark/endpoints/cluster_create.py
```diff
@@ -53,6 +53,11 @@ def execute(args: typing.NamedTuple):
 
     print_cluster_conf(cluster_conf)
 
+    spinner = utils.Spinner()
+
+    log.info("Please wait while your cluster is being provisioned")
+    spinner.start()
```
paselem: you probably want to put this into a catch / finally block since any crash in the middle should be handled and the stop would be called as expected.

cli/utils.py
```diff
@@ -1,3 +1,5 @@
+import threading
```
paselem: nitpick: alpha sort.

PR #278, opened 2017-12-14
Feature: Spark job submission

### LINKED ISSUE #219: Move logic for building spark-submit command to the node
Currently submitting a spark app (batch task) assumes that:
1. the cluster exists
2. the cluster has a master

Neither of these assumptions are necessary (or valid for job submission mode). To remove these assumptions, the logic that builds the spark-submit command should live on the node (in node_scripts/).

### LINKED ISSUE #247: node scripts are always uploaded to the same place in storage
Node scripts are always uploaded to blob container "spark-node-scripts" with the name "node-scripts.zip".

So if two pools are created with the same batch account, the most recent pool will overwrite the node scripts of the first pool.

### LINKED ISSUE #248: application resource files are uploaded by name only
Application files are uploaded by the application's name. This means that if an two apps are submitted with the same name, the resource files and/or output may be overwritten even if the apps are submitted to different clusters.

Applications should be uploaded by cluster/job id and with a generated value appended to them to avoid overwrites.

**Blob container**: cluster_id (or job_id)
**Blob Name**: \<app-name\>/\<filename\>

### LINKED ISSUE #277: Spark Job submission

### LINKED ISSUE #279: Cluster creation fails if spark-defaults.conf is not present
History server assumes spark-defaults.conf is present and if it is not, fails.

Should warn but not fail on missing spark-defaults.conf

### LINKED ISSUE #300: Start Task finishes before the node setup is completed
The start task should not exit until the node has been set up entirely.

This will prevent a race condition between app submissions and custom-scripts/requirements.

### LINKED ISSUE #329: Batch task exit code should match Spark application exit code

jafreck:
Fix #219 
Fix #248 
Fix #277 
Fix #279 
Fix #300 
Fix #329 
Fix #247


config/job.yaml
```diff
@@ -0,0 +1,55 @@
+# Job Configuration
+
```
jiata: I think we should add some comments here. things like:
- "an aztk spark job is a cluster an array of spark applications to run on that cluster ..."
- "an aztk spark job will automatically create/delete your cluster"

config/job.yaml
```diff
@@ -0,0 +1,55 @@
...
+job:
+    id: 
+    cluster_configuration:
```
jiata: can we put in some defaults? 
- vm size - d2_v2
- size - 5
- size low pri - 0
- docker_repo: [commented for smart default]
- custom_scripts: [commented]

config/job.yaml
```diff
@@ -0,0 +1,55 @@
...
+        vm_size: 
+        size: 
+        size_low_pri: 
+        docker_repo: 
+        custom_scripts: 
+
+    spark_configuration:
```
jiata: we can add the defaults `spark_defaults_conf: ".aztk/spark_defaults.conf"`
etc

config/job.yaml
```diff
@@ -0,0 +1,55 @@
...
+        spark_defaults_conf: 
+        spark_env_sh: 
+        core_site_xml: 
+  
+    applications:
```
jiata: we should specify that an application is the equivalent to a spark submit command as a comment

docs/70-jobs.md
```diff
@@ -0,0 +1,172 @@
...
+```sh
+aztk spark job list
+```
+
+
+### Viewing a Job
+To view details about a particular Job, run:
+
+```sh
+aztk spark Job get --id <your_job_id>
```
jiata: capital Job

aztk/client.py
```diff
@@ -159,6 +160,73 @@ def __get_remote_login_settings(self, pool_id: str, node_id: str):
...
+
+        # create job schedule and add task
+        setup = batch_models.JobScheduleAddParameter(
+            id=job_configuration.id,
+            schedule=schedule,
+            job_specification=job_spec)
+        
+        self.batch_client.job_schedule.add(setup)
+
+        return self.batch_client.job_schedule.get(job_schedule_id=job_configuration.id)
```
paselem: The method description says this returns None - should be updated

aztk/client.py
```diff
@@ -159,6 +160,73 @@ def __get_remote_login_settings(self, pool_id: str, node_id: str):
...
         return models.RemoteLogin(ip_address=result.remote_login_ip_address, port=str(result.remote_login_port))
 
+    def __submit_job(self, job_configuration, start_task, job_manager_task, autoscale_formula, software_metadata_key: str, vm_image_model):
+        """
+            Job Submission
+            :param job_configuration -> aztk_sdk.spark.models.JobConfiguration
+            :param start_task -> batch_models.StartTask
+            :param job_manager_task -> batch_models.TaskAddParameter
+            :param autoscale forumula -> str
+            :param software_metadata_key: str
```
paselem: all other params use '->' before the type. why ':' here?

aztk/client.py
```diff
@@ -159,6 +160,73 @@ def __get_remote_login_settings(self, pool_id: str, node_id: str):
         result = self.batch_client.compute_node.get_remote_login_settings(pool_id, node_id)
         return models.RemoteLogin(ip_address=result.remote_login_ip_address, port=str(result.remote_login_port))
 
+    def __submit_job(self, job_configuration, start_task, job_manager_task, autoscale_formula, software_metadata_key: str, vm_image_model):
```
paselem: nitpick: line too long

aztk/client.py
```diff
@@ -159,6 +160,73 @@ def __get_remote_login_settings(self, pool_id: str, node_id: str):
...
+
+        # define job specification
+        job_spec = batch_models.JobSpecification(
+            pool_info=batch_models.PoolInformation(auto_pool_specification=auto_pool_specification),
+            display_name=job_configuration.id,
+            on_all_tasks_complete=batch_models.OnAllTasksComplete.terminate_job,
+            job_manager_task=job_manager_task
+        )
+
+        # define schedule
```
paselem: Probably fine for now since we're not supporting anything but 'run now' but I would expect this is something that would be passed in as a parameter in the future.

aztk/spark/client.py
```diff
@@ -7,6 +7,7 @@
 from aztk.utils import helpers
 from aztk.spark.helpers import create_cluster as create_cluster_helper
 from aztk.spark.helpers import submit as submit_helper
```
paselem: minor: should submit_helper become cluster_submit_helper for clarity?

aztk/spark/client.py
```diff
@@ -121,7 +126,7 @@ def create_user(self, cluster_id: str, username: str, password: str = None, ssh_
 
     def get_application_log(self, cluster_id: str, application_name: str, tail=False, current_bytes: int = 0):
         try:
-            return get_log_helper.get_log(self.batch_client, cluster_id, application_name, tail, current_bytes)
+            return get_log_helper.get_log(self.batch_client, self.blob_client, cluster_id, application_name, tail, current_bytes)
```
paselem: nitpick: line too long

aztk/spark/client.py
```diff
@@ -131,3 +136,124 @@ def get_application_status(self, cluster_id: str, app_name: str):
...
+                                                                 job_configuration.spark_configuration)
+            
+            start_task = create_cluster_helper.generate_cluster_start_task(self,
+                                                                           zip_resource_files,
+                                                                           job_configuration.gpu_enabled,
+                                                                           job_configuration.docker_repo)
+             
+            application_tasks = []
+            for application in job_configuration.applications:
+                application_tasks.append((application, submit_helper.generate_task(self, job_configuration.id, application)))
```
paselem: nitpick: lots of lines too long in this file.

aztk/spark/client.py
```diff
@@ -131,3 +136,124 @@ def get_application_status(self, cluster_id: str, app_name: str):
...
+                autoscale_formula = "maxNumberofVMs = {0}; targetNumberofVMs = {1}; $TargetDedicatedNodes=min(maxNumberofVMs, targetNumberofVMs)".format(
+                    job_configuration.max_dedicated_nodes, job_configuration.max_dedicated_nodes)
+            elif job_configuration.max_low_pri_nodes and not job_configuration.max_dedicated_nodes:
+                autoscale_formula = "maxNumberofVMs = {0}; targetNumberofVMs = {1}; $TargetLowPriorityNodes=min(maxNumberofVMs, targetNumberofVMs)".format(
+                    job_configuration.max_low_pri_nodes, job_configuration.max_low_pri_nodes)
+            else:
+                raise error.AztkError("Jobs do not support both dedicated and low priority nodes. JobConfiguration fields max_dedicated_nodes and max_low_pri_nodes are mutually exclusive values.")
+            
+            
+            job = self.__submit_job(
```
paselem: Does it make sense to return the job ID for monitoring purposes?

aztk/spark/client.py
```diff
@@ -131,3 +136,124 @@ def get_application_status(self, cluster_id: str, app_name: str):
...
+
+
+            software_metadata_key = "spark"
+
+            vm_image = models.VmImage(
+                publisher='Canonical',
+                offer='UbuntuServer',
+                sku='16.04')
+            
+            if job_configuration.max_dedicated_nodes and not job_configuration.max_low_pri_nodes:
```
paselem: Do you need this logic? Can't you always just set them both but one will always be 0?

aztk/spark/client.py
```diff
@@ -131,3 +136,124 @@ def get_application_status(self, cluster_id: str, app_name: str):
...
+                autoscale_formula=autoscale_formula,
+                software_metadata_key=software_metadata_key,
+                vm_image_model=vm_image)
+        
+        except batch_error.BatchErrorException as e:
+            raise error.AztkError(helpers.format_batch_exception(e))
+
+    def list_jobs(self):
+        try:
+            return [models.Job(cloud_job_schedule) for cloud_job_schedule in self.batch_client.job_schedule.list()]
```
paselem: Is this correct? You're asking to list jobs, but returning job schedules... Is the expectation that you can get a job from a schedule elsewhere?

aztk/spark/client.py
```diff
@@ -131,3 +136,124 @@ def get_application_status(self, cluster_id: str, app_name: str):
...
+        except batch_error.BatchErrorException as e:
+            raise error.AztkError(helpers.format_batch_exception(e))
+    
+    def stop_job(self, job_id):
+        try:
+            return job_submit_helper.stop(self, job_id)
+        except batch_error.BatchErrorException as e:
+            raise error.AztkError(helpers.format_batch_exception(e))
+
+    # def enable_job(self, job_id):
```
paselem: Remove or uncomment code?

aztk/spark/client.py
paselem: Also, curious as far as coding pattern goes, that all the other methods proxy through job_submit_helper.foo() while this one hits the batch API directly. Any reason in particular?

aztk/spark/helpers/create_cluster.py
```diff
@@ -23,13 +23,13 @@ def __docker_run_cmd(docker_repo: str = None, gpu_enabled: bool = False, file_mo
         cmd = CommandBuilder('docker run')
     cmd.add_option('--net', 'host')
     cmd.add_option('--name', constants.DOCKER_SPARK_CONTAINER_NAME)
-    cmd.add_option('-v', '/mnt/batch/tasks:/batch')
+    cmd.add_option('-v', '/mnt/batch/tasks:/mnt/batch/tasks')
```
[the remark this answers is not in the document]
paselem: If we go forward with it, we may want to expose both /mnt/batch/tasks and /batch for a while for backwards compat.

aztk/spark/helpers/job_submission.py
```diff
@@ -0,0 +1,144 @@
...
+import time
+import azure.batch.models as batch_models
+from aztk.utils import constants, helpers
+from aztk.utils.command_builder import CommandBuilder
+
+'''
+    Job Submission helper methods
+'''
+def __app_cmd():
+    # TODO: convert to CommandBuilder
```
paselem: Are you going to implement this TODO?

aztk/spark/helpers/job_submission.py
```diff
@@ -0,0 +1,144 @@
...
+    return 'sudo docker exec -i -e AZ_BATCH_TASK_WORKING_DIR=$AZ_BATCH_TASK_WORKING_DIR -e AZ_BATCH_JOB_ID=$AZ_BATCH_JOB_ID spark /bin/bash >> output.log 2>&1 -c "python \$DOCKER_WORKING_DIR/job_submission.py"'
+
+def generate_task(spark_client, job, application_tasks):
+    # Upload dependent JARS
```
paselem: Is this comment valid? I don't see where the jars are.

aztk/spark/helpers/job_submission.py
```diff
@@ -0,0 +1,144 @@
...
+    # Create task
+    task = batch_models.JobManagerTask(
+        id=job.id,
+        command_line=helpers.wrap_commands_in_shell([task_cmd]),
+        resource_files=resource_files,
+        kill_job_on_completion=False, 
+        allow_low_priority_node=True,
+        user_identity=batch_models.UserIdentity(
+            auto_user=batch_models.AutoUserSpecification(
+                scope=batch_models.AutoUserScope.task,
```
paselem: Hrm... will this scope cause issues later on? For example if everything in the cluster is create w/ PoolAdmin permissions (like a file) will TaskAdmin have access to it?

aztk/client.py
jafreck: I was debating whether this should return anything. I think for consistency with cluster_create, this should probably return the azure.batch.models.CloudJobSchedule like it's doing, and then the aztk.spark client should convert that to a aztk.spark.models.Job. 

i'll update the method description and aztk.spark.client to reflect that.

aztk/spark/helpers/job_submission.py
```diff
@@ -0,0 +1,144 @@
...
+
+    # recent_run_job = __get_recent_job(spark_client, job_id)
+    # return spark_client.get_application_log(recent_run_job.id.replace(":", "-"), application_name)
+    return spark_client.get_application_log(job_id, application_name)
+
+
+def stop_app(spark_client, job_id, application_name):
+    recent_run_job = __get_recent_job(spark_client, job_id)
+
+    # TODO: stop spark job on node
```
paselem: Why do we need to manually stop the spark job on the node? Won't that happen automatically if we kill the driver app (i.e. the task)?

aztk/spark/helpers/job_submission.py
```diff
@@ -0,0 +1,144 @@
...
+    # TODO: stop spark job on node 
+
+    # stop batch task
+    try:
+        spark_client.batch_client.task.terminate(job_id=recent_run_job.id, task_id=application_name)
+        return True
+    except batch_models.batch_error.BatchErrorException:
+        return False
+
+def wait_until_job_finished(spark_client, job_id):
```
paselem: I think we've had this conversation in the past, but I really don't like methods like this built into the core SDK - it should happen at a higher level (like the CLI or calling client). There is little to no monitoring here and a caller can't control most of the polling logic.

aztk/spark/helpers/submit.py
```diff
@@ -1,5 +1,7 @@
 from typing import List
-
+import os
```
paselem: nitpick: alpha sort.

aztk/spark/helpers/submit.py
```diff
@@ -13,79 +15,11 @@ def __get_node(spark_client, node_id: str, cluster_id: str) -> batch_models.Comp
...
-        docker_exec_cmd.to_str()
-    ]
-
-
-def submit_application(spark_client, cluster_id, application, wait: bool = False):
-    """
-    Submit a spark app
-    """
-
+def generate_task(spark_client, container_id, application):
```
paselem: this looks very similar to the generate_task method in aztk/spark/helpers/job_submission.py. Can we only write it once?

aztk/spark/models.py
```diff
@@ -125,11 +125,92 @@ def __init__(
...
+class Job():
+    def __init__(self, cloud_job_schedule: batch_models.CloudJobSchedule, cloud_tasks: List[batch_models.CloudTask] = None):
+        self.id = cloud_job_schedule.id
+        self.last_modified = cloud_job_schedule.last_modified
+        self.state = cloud_job_schedule.state._value_
+        self.state_transition_time = cloud_job_schedule.state_transition_time
+        self.creation_time = cloud_job_schedule.creation_time
+        self.applications = [Application(task) for task in (cloud_tasks or [])] 
+
+        # self.schedule = cloud_job_schedule.schedule
```
paselem: remove commented code

aztk/utils/helpers.py
```diff
@@ -60,7 +60,32 @@ def wait_for_task_to_complete(job_id: str, task_id: str, batch_client):
             return
 
 
-def upload_file_to_container(container_name, file_path, blob_client=None, use_full_path=False, node_path=None) -> batch_models.ResourceFile:
+def upload_text_to_container(container_name: str, application_name: str, content: str, file_path: str, blob_client=None) -> batch_models.ResourceFile:
+    blob_name = file_path
+    blob_path = application_name + '/' + blob_name # + '/' + time_stamp + '/' + blob_name
```
paselem: remove commented code.

cli/spark/endpoints/job/get_app_logs.py
```diff
@@ -0,0 +1,29 @@
...
+
+    # parser.add_argument('--tail', dest='tail', action='store_true')
+
+
+def execute(args: typing.NamedTuple):
+    spark_client = load_spark_client()
+    app_logs = spark_client.get_job_application_log(args.job_id, args.app_name)
+    print(app_logs.log)
+
+    # if args.tail:
```
paselem: remove? (same w/ line 17)

cli/spark/endpoints/job/submit.py
```diff
@@ -0,0 +1,65 @@
...
+                        dest='job_conf',
+                        required=False,
+                        help='Path to the job.yaml configuration file. Defaults to .aztk/job.yaml')
+
+
+def execute(args: typing.NamedTuple):
+    spark_client = load_spark_client()
+    job_conf = JobConfig()
+
+    job_conf.merge(args.job_id, args.job_conf)
```
paselem: probably a good idea to check that job_conf is valid

cli/utils.py
```diff
@@ -223,6 +224,84 @@ def print_batch_exception(batch_exception):
...
+        log.info(
+            print_format.format(
+                job.id,
+                job.state,
+                utc_to_local(job.creation_time)
+            )
+        )
+
+def utc_to_local(utc_dt):
+    return utc_dt.replace(tzinfo=datetime.timezone.utc).astimezone(tz=None).strftime("%H:%M%p %d/%m/%y")
```
paselem: nitpick: this method shouldn't be nested between job methods

nitpick2: Remove commented lines in print_job method

docs/50-sdk.md
```diff
@@ -223,10 +224,204 @@ Find some samples and getting stated tutorial in the `examples/sdk/` directory o
...
+    Submit an AZTK Spark Job
+
+    Parameters:
+
+        - job_configuration: aztk.spark.models.JobConfiguration
+            The configuration of the job to be submitted
+    
+    Returns:
+        
+        - None
```
paselem: Update this return value

docs/50-sdk.md
```diff
@@ -223,10 +224,204 @@ Find some samples and getting stated tutorial in the `examples/sdk/` directory o
...
+    List all created AZTK Spark Jobs
+
+    Parameters:
+
+        - job_configuration: aztk.spark.models.JobConfiguration
+            The configuration of the job to be submitted
+    
+    Returns:
+        
+        - List[aztk.spark.models.Job]
```
paselem: So... this again is probably fine for now but once we introduce schedules what is this going to return? The list of jobs or the schedule? Or maybe the list of running and completed jobs and the time when it will run again?

docs/50-sdk.md
```diff
@@ -223,10 +224,204 @@ Find some samples and getting stated tutorial in the `examples/sdk/` directory o
...
+    Parameters:
+
+        - job_id: str
+            The id of the Job
+    
+    Returns:
+        
+        - List[aztk.spark.models.Application]
+       
+- `get_job(self, job_id)`
```
paselem: Similar to comment above... We created a schedule, but maybe have no way to reference it or modify it.

docs/70-jobs.md
```diff
@@ -0,0 +1,172 @@
...
+Once submitted, this Job will run two applications, pipy100 and pipy200, on an automatically provisioned Cluster with 3 dedicated Standard_f2 size Azure VMs. Immediately after both pipy100 and pipy200 have completed the Cluster will be destroyed. Application logs will be persisted and available.
+
+### Commands
+Submit a Spark Job:
+
+```sh
+aztk spark job submit --id <your_job_id> --job-conf </path/to/job.yaml>
+```
+
+NOTE: The Job id (`--id`) can only contain alphanumeric characters including hyphens and underscores, and cannot contain more than 64 characters. Each cluster **must** have a unique cluster id.
```
paselem: Do the application names have the same requirement? I believe they do - if so, worth mentioning here.

docs/70-jobs.md
```diff
@@ -0,0 +1,172 @@
...
+
+### Get information about a Job's Application
+To get information about a Job's Application:
+
+```sh
+aztk spark job get-app --id <your_job_id> --name <your_application_name>
+```
+
+
+### Getting a Job's Application's log
```
paselem: Is there a way to delete application logs? If not we should have an issue for that after this ships.

node_scripts/submit.py
```diff
@@ -0,0 +1,172 @@
...
+    spark_submit_cmd.add_option('--driver-memory', driver_memory)
+    spark_submit_cmd.add_option('--executor-memory', executor_memory)
+    spark_submit_cmd.add_option('--driver-cores', driver_cores)
+    spark_submit_cmd.add_option('--executor-cores', executor_cores)
+
+    spark_submit_cmd.add_argument(
+        os.environ['AZ_BATCH_TASK_WORKING_DIR'] + '/' + app + ' ' +
+        ' '.join(['\'' + app_arg + '\'' for app_arg in app_args if app_args]))
+
+    # print("SPARK SUBMIT CMD: ", spark_submit_cmd.to_str())
```
paselem: remove?

node_scripts/wait_until_master_selected.py
```diff
@@ -0,0 +1,20 @@
+
```
paselem: nitpick: remove blank lines

aztk/client.py
jafreck: Originally, the JobConfiguration model had a fields for each of these values. Since we are going with just 'run now' for the time being, I removed those. 

When we add scheduling functionality, we should probably add those fields (or an abstraction layer for those fields) back to the JobConfiguration model.

aztk/spark/client.py
jafreck: For the CLI, I have validation in place for when the file is read. However, for the SDK, I'm not sure where the best place to put the validation is. I agree that this method probably shouldn't have logic to validate the job_configuration objection. But, I'm not sure where the best place to validate is. Perhaps on the model initialization. What do you think?

aztk/spark/helpers/job_submission.py
```diff
@@ -0,0 +1,144 @@
...
+        user_identity=batch_models.UserIdentity(
+            auto_user=batch_models.AutoUserSpecification(
+                scope=batch_models.AutoUserScope.task,
+                elevation_level=batch_models.ElevationLevel.admin))
+    )
+
+    return task
+
+
+def __get_recent_job(spark_client, job_id):
```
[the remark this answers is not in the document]
jafreck: this is just mimicking the batch service terminology. a  Batch job schedule has a "recent_job" which is the most recent Batch job that has run under the job_schedule.

aztk/spark/helpers/submit.py
jafreck: I think that unfortunately, they are sufficiently different to require two different methods. 

This one takes aztk applications, uploads the necessary resource files and creates Batch task objects from them. 

The generate_task in job_submission.py takes an array of aztk application definitions, serializes the definitions and uploads the serialized object as a resource file for a the job manager task it creates.

docs/50-sdk.md
jafreck: For list in particular, I think that this should probably still return AZTK Jobs. When we add scheduling, I think the AZTK Job model should be updated to include a list of all Batch Jobs that have run under it.

The CLI on list should display display:
1. The name of the JobSchedule
2. The state of the job schedule
3. The Job Schedule's creation time

On a get, the CLI should display all of the above plus the state of all Batch Jobs that have run under the Job Schedule.

We should provide the ability to get information & logs about the applications for each of those runs (batch jobs).

docs/50-sdk.md
jafreck: That's by design for now. At this stage, for job submission, there isn't really a schedule. (technically there is under the hood, but it is not exposed to the user yet). So it doesn't really make sense to have an update to an AZTK Job, becuase AZTK jobs run exactly once the instant they are created.

Once we have scheduling we should definitely add an update/modify AZTK job method.

docs/70-jobs.md
jafreck: Not currently. Will open an issue. Cleaning up storage should also happen as a part of cluster deletion for interactive mode like we have talked about before.

aztk/spark/helpers/job_submission.py
jafreck: I'm not entirely sure on this, I'll have to test.

jiata:
Wanted to revive this PR. Is this waiting on @paselem's review?

jafreck:
This is waiting on a couple changes and better error handling. I'll be updating it today.

cli/utils.py
```diff
@@ -223,6 +224,77 @@ def print_batch_exception(batch_exception):
...
+
+def print_job(job: aztk.spark.models.Job):
+    print_format = '{:<36}| {:<15}'
+
+    log.info("")
+    log.info("Job             %s", job.id)
+    log.info("------------------------------------------")
+    log.info(print_format.format("State:", job.state))
+    log.info(print_format.format("Transition Time:", utc_to_local(job.state_transition_time)))
+    log.info("")
```
paselem: Nitpick: This doesn't follow the format for print cluster:
image


vs.

image

cli/utils.py
```diff
@@ -223,6 +224,77 @@ def print_batch_exception(batch_exception):
...
+
+    print_format = '{:<36}| {:<15}| {:<14}'
+    print_format_underline = '{:-<36}|{:-<16}|{:-<17}'
+    log.info(print_format.format("Applications", "State", "Transition Time"))
+    log.info(print_format_underline.format('', '', '', ''))
+
+    for application in job.applications:
```
paselem: Before printing applications, I think we should call the print_cluster method here (or some summary of it). It's incredibly valuable to see it inline at this point, especially when the cluster is being provisioned and the applications appear to be waiting or stuck.

jiata:
will have to add max_retry_count in .aztk/job.yaml

jiata:
When you `get-app` of a task that doesn't yet exist, you get an error that could be confusing:

screen shot 2018-01-10 at 11 14 28 am

I think this is confusing because this implies that the task/app (maybe we should also keep customer-facing names consistent...) has not been created, when to the user, it should have been created already. Instead, maybe we can throw them an msg along the lines of: "your app exists, but no data can be retrieved as the app has not yet started"

jiata:
I'm getting an error when running the `spark job get` cmd:

screen shot 2018-01-10 at 11 18 46 am

based on the portal, my cluster is up and running. One node is in 'running' state, the other is in 'waiting for start task' state.

jiata:
can you also remind me on what we decided with monitoring a live job? did we decide to table that feature for a different PR?

jiata:
`get-app-logs` cmd will basically not show anything until the app in spark starts running, right? this *could* be a bad experience... For a user who runs this command shortly after submitting the aztk job, they will basically have no feedback until the job kicks off (which will likely be in the 5 minutes + range with cluster provision time). 

Can we add a message along the lines of "pending... your job is being set up, which could take as long as 5 minutes"

paselem:
@jiata  - I agree, and is somewhat in line w/ my comment above. We need to make the cluster provisioning state quite obvious, especially if it means that it will delay the rendering of tasks. It will probably be a bit strange to users at first that the tasks (apps!) won't actually get created until the cluster is up and running.

jiata:
I'm getting this msg when i run the `get-app-logs` cmd. 

screen shot 2018-01-10 at 11 25 07 am

To me, this is quite confusing since my job *was* created and *was not* deleted.

jiata:
another point regarding `get-app-logs` cmd:

Is there a way the command can take a parameter like `spark job get-app-logs --output path/to/folder/to/dump/log/file`? Since a lot of the logs will likely be thousands or even tens of thousands of lines long, having the option to dump out log files would be very useful.

paselem:
I also ran into the issue w/ aztk not finding my logs. It turned out that the app ran successfully, but the logs were not uploaded to the storage account. I think it's a bug.

jiata:
@paselem Well the `get-app-logs` only pulls from storage right? and we persist the logs to storage only after the app/task is completed. So `get-app-logs` shouldn't work until after completed. @jafreck - is that correct? 

Regardless, we dont really give the user much insight into what actually is happening, so it may just appear broken to them.

jiata:
can we add a `get-apps` command (in addition to the `get-app` command that we already have)
```
$ aztk spark job get-apps --id my_job

App Name      State            State Transition Time 
app3          active      11:39 AM 1/1/1
app2          completed   11:32 AM 1/1/1
app1          completed   11:31 AM 1/1/1 
```

paselem:
@jiata  - we should probably call it 'list-apps', but yeah, we could.

config/job.yaml
```diff
@@ -0,0 +1,59 @@
+# Job Configuration
+#   An Aztk Job is a cluster and an array of Spark applications to run on that cluster
+#   AZTK Spark Jobs will automatically manage the lifecycle of the cluster
```
jiata: would it make sense to add a link to the docs/70-jobs.md here?

cli/utils.py
```diff
@@ -249,33 +249,98 @@ def print_job(client, job: aztk.spark.models.Job):
...
+
+def print_cluster_summary(cluster: aztk.spark.models.Cluster):
+    print_format = '{:<4} {:<23} {:<15}'
+
+    log.info("Cluster           %s", cluster.id)
+    log.info("-"*42)
+    log.info("Nodes             %s", __pretty_node_count(cluster))
+    log.info("| Dedicated:      %s", __pretty_dedicated_node_count(cluster))
+    log.info("| Low priority:   %s", __pretty_low_pri_node_count(cluster))
+    state_count = node_state_count(cluster)
```
paselem: One last change I'd add here: If the master is not yet available, I would show that, and then in the application_summary print method I would make a statement that no applications are scheduled until a master node is assigned.

jafreck: Fixed. Now, if the master is not selected a warning is logged below the job and application summaries.

PR #282, opened 2017-12-15
Bug: fix loading local spark config

### LINKED ISSUE #281: load_spark_config fails to find non-global configuration files
This causes an issue with history server that crashes the container.

jafreck:
Fix #281


jiata:
@jafreck is this change reflected in stable branch too?

jafreck:
yes, it is now. I updated stable and the v0.5.0 tag to have this change.

PR #291, opened 2017-12-19
jupyter azfiles bug + gpu sample

jiata:
- move mount point for jupyter custom script to /mnt (instead of /jupyter)
- move samples to /mnt
- created a GPU sample with numba in jupyter


jafreck:
It feels a little weird to have the samples in `/mnt`. Can we keep the samples in `/jupyter`?

jiata:
the reason to put it in /mnt is because this is where people will mount their file-share. the recommendation for using azure files is to persist their notebooks. 

Without jupyter running from /mnt, users will not be able to access their mnt point, and will have to use the command line to move files to and from their share.

PR #303, opened 2017-12-23
Feature: on node user creation

### LINKED ISSUE #272: Move user creation to on node
If we move the code to create a user from the client to one of the nodes, then `aztk cluster create` would no longer need to wait for the cluster to be created, it could just immediately return.

### LINKED ISSUE #305: UI Bug when prompted for a password while cluster is being created
The spinner interferes with the password prompt.

```
\Please input a password for user 'spark':|
Please confirm your password for user 'spark':|
```

jafreck:
Fix #272
Fix #305


aztk/spark/client.py
```diff
@@ -20,7 +21,14 @@ def __init__(self, secrets_config):
     '''
     def create_cluster(self, cluster_conf: models.ClusterConfiguration, wait: bool = False):
         try:
-            zip_resource_files = upload_node_scripts.zip_scripts(self.blob_client, cluster_conf.custom_scripts, cluster_conf.spark_configuration)
+            if cluster_conf.user_configuration:
+                user_conf = yaml.dump({'username': cluster_conf.user_configuration.username,
+                                       'password': cluster_conf.user_configuration.password,
```
paselem: We may want to omit password until we support encryption.

aztk/spark/client.py
```diff
@@ -20,7 +21,14 @@ def __init__(self, secrets_config):
...
+                                       'ssh-key': cluster_conf.user_configuration.ssh_key,
+                                       'cluster_id': cluster_conf.cluster_id})
+            else:
+                user_conf = None
+            zip_resource_files = upload_node_scripts.zip_scripts(self.blob_client, cluster_conf.custom_scripts, cluster_conf.spark_configuration, user_conf)
```
paselem: nitpick: line too long

aztk/spark/utils/upload_node_scripts.py
```diff
@@ -114,5 +114,9 @@ def zip_scripts(blob_client, custom_scripts, spark_conf):
             for jar in spark_conf.jars:
                 zipf = __add_file_to_zip(zipf, jar, 'jars', binary=True)
 
+    if user_conf:
+        zipf = __add_str_to_zip(zipf, user_conf, 'user.yaml')
```
paselem: why _add_str_to_zip? Isn't this a file?

jafreck: The user_conf is a byte string that is written to a file in node_scripts called `user.yaml`. `user.yaml` is not a file on the client. 

__add_str_to_zip() is probably not the best name for what this method does, though.

aztk/spark/client.py
jafreck: If we disable password here that means that we would no longer support adding a user with password at cluster creation time. That's a potentially breaking change, so maybe this should PR should wait for or include the encryption feature.

jafreck: If the goal is just to not have the password in cleartext, we could also hash it, and create the user using the hash. 

That would break ssh'ing outside of AZTK, though.

aztk/models.py
```diff
@@ -23,6 +23,13 @@ def __init__(self, name: str = None, script: str = None, run_on=None):
         self.run_on = run_on
 
 
+class UserConfiguration:
+    def __init__(self, username: str, ssh_key: str = None, password: str = None):
```
paselem: Will this get auto-populated from the users' secrets.yaml?

jafreck: On the CLI, yeah -- that happens here: 
https://github.com/jafreck/aztk/blob/06f85c2538db98e168fed8c38fe9b88524f95e5f/cli/spark/endpoints/cluster/cluster_create.py#L87

For the SDK, no the secrets.yaml file is not used at all.

PR #304, opened 2017-12-27
Feature: Cluster Run and Copy

### LINKED ISSUE #192: Installing packages on the container are not recognized when running submit
Caused by: https://github.com/Azure/aztk/blob/feature/new_containers/docker-image/python/spark2.2.0/Dockerfile#L12

### LINKED ISSUE #220: Add support for cluster-wide commands
Add a `aztk spark cluster execute` command that allows users to submit commands to the cluster.

Useful for things like setting software specific environment variables in interactive mode.

### LINKED ISSUE #280: Add command cluster scp

jafreck:
Fix #280 
Fix #220 
Fix #192


aztk/client.py
```diff
@@ -159,13 +178,69 @@ def __get_remote_login_settings(self, pool_id: str, node_id: str):
...
+                self.__create_user(pool_id=pool_id, node_id=node_id, username=username, ssh_key=ssh_key)
+            except batch_error.BatchErrorException as error:
+                print(error)
+                raise error
+        return ssh_key
+
+    def __create_user_on_pool(self, username, pool_id, nodes):
+        ssh_key = RSA.generate(2048)
+        with concurrent.futures.ThreadPoolExecutor() as executor:
+            futures = {executor.submit(self.__create_user_on_node, username, pool_id, node.id, ssh_key.publickey().exportKey('OpenSSH').decode('utf-8')): node for node in nodes}
```
paselem: nitpick: line too long (a few of these around)

aztk/client.py
```diff
@@ -159,13 +178,69 @@ def __get_remote_login_settings(self, pool_id: str, node_id: str):
...
+            concurrent.futures.wait(futures)
```
paselem: Should this have a retry policy? What happens if we cannot connect to a node?

aztk/client.py
```diff
@@ -159,13 +178,69 @@ def __get_remote_login_settings(self, pool_id: str, node_id: str):
...
+            futures = [exector.submit(self.__delete_user, pool_id, node.id, username) for node in nodes]
+            concurrent.futures.wait(futures)
+
+
+    def __cluster_run(self, cluster_id, container_name, command):
+        pool, nodes = self.__get_pool_details(cluster_id)
+        nodes = [node for node in nodes]
+        cluster_nodes = [self.__get_remote_login_settings(pool.id, node.id) for node in nodes]
+        try:
+            ssh_key = self.__create_user_on_pool('aztk', pool.id, nodes)
```
paselem: How long can this take?

aztk/client.py
```diff
@@ -159,13 +178,69 @@ def __get_remote_login_settings(self, pool_id: str, node_id: str):
...
+            asyncio.get_event_loop().run_until_complete(ssh_lib.clus_exec_command(command,
+                                                                                  container_name,
+                                                                                  'aztk',
+                                                                                  cluster_nodes,
+                                                                                  ssh_key=ssh_key.exportKey().decode('utf-8')))
+            self.__delete_user_on_pool('aztk', pool.id, nodes)
```
paselem: shouldn't the delete user code be in a finally clause? If anything fails in between the create and fail, it could leave lingering accounts around

aztk/client.py
```diff
@@ -159,13 +178,69 @@ def __get_remote_login_settings(self, pool_id: str, node_id: str):
...
+        except OSError as exc:
+            raise exc
+
+    def __cluster_copy(self, cluster_id, container_name, source_path, destination_path):
+        pool, nodes = self.__get_pool_details(cluster_id)
+        nodes = [node for node in nodes]
+        cluster_nodes = [self.__get_remote_login_settings(pool.id, node.id) for node in nodes]
+        try:
+            ssh_key = self.__create_user_on_pool('aztk', pool.id, nodes)
```
paselem: Would it be OK to either use an existing cluster user (if we know it) OR just keep the user around assuming it was created securely?

aztk/utils/ssh.py
```diff
@@ -0,0 +1,98 @@
...
+    )
+
+    return client
+
+
+def node_exec_command(command, container_name, username, hostname, port, ssh_key=None, password=None):
+    client = connect(hostname=hostname, port=port, username=username, password=password, pkey=ssh_key)
+    docker_exec = 'sudo docker exec 2>&1 -t {0} /bin/bash -c \'set -e; set -o pipefail; {1}; wait\''.format(container_name, command)
+    stdin, stdout, stderr = client.exec_command(docker_exec, get_pty=True)
+    [print(line.decode('utf-8')) for line in stdout.read().splitlines()]
```
paselem: should you also print stderr?

cli/spark/endpoints/cluster_run.py
```diff
@@ -0,0 +1,18 @@
...
+                        dest='cluster_id',
+                        required=True,
+                        help='The unique id of your spark cluster')
+    parser.add_argument('command',
+                        help='The command to run on your spark cluster')
+
+def execute(args: typing.NamedTuple):
+    spark_client = load_spark_client()
+    result = spark_client.cluster_run(args.cluster_id, args.command)
+    #TODO: pretty print result
```
paselem: I think this is worth doing before merging.

cli/spark/endpoints/cluster_copy.py
```diff
@@ -0,0 +1,28 @@
...
+from cli import utils
+from cli.spark.aztklib import load_spark_client
+import aztk.spark
+
+
+def setup_parser(parser: argparse.ArgumentParser):
+    parser.add_argument('--id', dest='cluster_id', required=True,
+                        help='The unique id of your spark cluster')
+
+    parser.add_argument('--source-path', required=True,
```
paselem: can/should this include directories? (happy to push that out to later)

jafreck: yeah -- it is just an sftp client under the hood so directories will work fine.

cli/spark/endpoints/cluster_run.py
jafreck: Not sure how this is best done.

Currently, the output of the command is streaming from each node and printed as it arrives.

There are a couple options to clean this up. First is to wait to print until the entire output of the command is received. Then print the entire output of the command in a single block for each node. This has the advantage of being readable, but provides no feedback to the user until the entire command has completed (will look like it is hanging until then). This could be confusing for commands that provides feedback while they are executing.

The other option is to only print the output of the command from one node. Most of the time, this will be fine. But if there is an error on one node, It's not clear how we would convey that.

Ideally, we could stream the output from each node, and output them in blocks as they come, but I don't think that is possible (at least I'm not sure how it would be done).

aztk/utils/ssh.py
jafreck: in line 45, I'm redirecting stdout and stderr so they are returned together

aztk/client.py
jafreck: It would be fine to do, but users shouldn't have to care about users to run commands or copy files. 

In general, the entire concept of a user is not really useful to users and can even be confusing. For example, when I ssh in, I specify a user, but regardless of what user is specified, I'm logged into the container as root.

aztk/client.py
jafreck: about 1 second overhead regardless of number of nodes in the cluster since they are created concurrently.

PR #320, opened 2018-01-08
Feature: expose exit_code

### LINKED ISSUE #314: app-logs should return exit code
Currently, there is no easy way to detect if a spark job submitted to the cluster has completed successfully. Please print/return exit code 0, if the job has completed successfully.

jafreck:
Fix #314


jafreck:
@paselem @jiata Should this be extended to the CLI as well? Or is the CLI printing the entire Spark log sufficient?

paselem:
@jafreck - what would this look like in the cli output? I would think that having this in the "list view" of applications is valuable.

jafreck:
I might be misunderstanding what you mean, but we don't have a "list applications" view on the CLI. 

The CLI has an app-logs command, but that prints the entire spark-submit log to the console. I didn't feel like printing the return code belonged there (which is why I originally left it out). 

For job submission, adding the exit code to the application list would be a nice feature. But for cluster mode, I don't think it fits.

paselem:
I'm referring to this view:

image

jafreck:
That view is part of the Job submission branch which is still out for PR itself. I can update that branch once this change is in master.

PR #321, opened 2018-01-08
Feature: WASB performance tune core-site.xml

### LINKED ISSUE #319: Speed up WASB reads
about 30% to 40% wasb read perf gain by tuning this parameter:

Slow reading speed from wasb in docker image compared agaisnt HDI's wasb config. HDI has  “io.file.buffer.size” configured which can affect the reading speed.
 
In aztk, we just use default spark package. In this case, “io.file.buffer.size” is not set and default value 4096 will be applied. But in HDI cluster, they set the value to 131072. This value can speed up the reading by about about 30% to 40%!
  
How to apply the new value? Basically, just need to add/modify core-site.xml file in $SPARK_HOME/conf folder with these context:
```
<?xml version="1.0" encoding="UTF-8"?>
<?xml-stylesheet type="text/xsl" href="configuration.xsl"?>
<configuration>
    <property>
      <name>io.file.buffer.size</name>
      <value>131072</value>
    </property>
</configuration>
```

jafreck:
Fix #319

### DIFF
--- a/config/core-site.xml
+++ b/config/core-site.xml
@@ -2,6 +2,11 @@
 <?xml-stylesheet type="text/xsl" href="configuration.xsl"?>
 <configuration>
 
+  <property>
+    <name>io.file.buffer.size</name>
+    <value>131072</value>
+  </property>
+
 <!-- WASB (Azure Storage) Configuration  -->
 <!-- More information about WASB can be foudn at https://docs.microsoft.com/azure/hdinsight/hdinsight-hadoop-use-blob-storage> -->
 <!-- <property>


config/core-site.xml
```diff
@@ -2,6 +2,11 @@
 <?xml-stylesheet type="text/xsl" href="configuration.xsl"?>
 <configuration>
 
+  <property>
```
paselem: I would comment this out and add it next to the WASB configuration and/or the ADLS configuration.

jafreck:
From what I've read, this is a setting that is valuable outside of just WASB use. The common suggestion is to set it to 64k or 128k (what it is at now), and just leave it. This gives an across the board performance boost on modern systems.

PR #324, opened 2018-01-11
Feature: VNet support 

### LINKED ISSUE #275: Allow users to create a cluster with an existing VNET. It should not be removed as part of the cluster delete command.

timotheeguerin:
fix #275


config/secrets.yaml.template
```diff
@@ -1,15 +1,22 @@
...
-batch:
-    batchaccountname: 
-    batchaccountkey: 
-    batchserviceurl: 
 
-storage:
-    storageaccountname: 
-    storageaccountkey: 
-    storageaccountsuffix: core.windows.net
+servicePrincipal:
```
jafreck: Should be changed to serviceprincipal for consistency. (in general, this file should probably have underscores to separate words like cluster.yaml does, but that is a breaking change so I'm not sure we want to go through with that).

config/secrets.yaml.template
```diff
@@ -1,15 +1,22 @@
...
+    tenantid: 
+    clientid: 
+    credential: 
+    batchaccountresourceid: 
+    storageaccountresourceid: 
+
+sharedKey:
```
jafreck: Should be sharedkey

jafreck:
Also, there isn't any change to cluster.yaml in config/. That should be updated to include a commented out example of subnet_id. 

I feel like subnet_id might be better named as subnet_resource_id, but if we ship with a commented out example we are probably fine.

config/secrets.yaml.template
timotheeguerin: well this is already a breaking change this new format so we could go for it

timotheeguerin:
yeah we are missing docs for VNET, I'll work on that

config/secrets.yaml.template
jafreck: I don't think this is a breaking change. We will still support the existing format, and the new service principal format.

If we only add underscores for the service principal format, we wouldn't break anything but then we have a inconsistent syntax within the file.

cli/config.py
```diff
@@ -39,19 +45,62 @@ def _load_secrets_config(self, path: str=aztk.utils.constants.DEFAULT_SECRETS_PA
...
+            if not self.service_principal_client_id:
+                raise aztk.error.AztkError("Service principal auth missing client_id")
+            if not self.service_principal_credential:
+                raise aztk.error.AztkError("Service principal auth missing credential")
+        else:
+            if self.service_principal_tenant_id or self.service_principal_client_id or \
+               self.service_principal_credential:
+                raise aztk.error.AztkError(
+                    "Service principal auth incomplete: must have tenantid, clientid, credential "
+                    "and at least one of batchaccountresourceid and storageaccountresourceid.")
```
timotheeguerin: here shouldn't the message be updated? batch_account resource id is required right?

timotheeguerin:
@emlyn I restructured the `SecretsConfiguration` to have a better separation between service principal and shared key and follow the yaml file sturcutre. Everything seems to be working, if you could try on your side too to double check I didn't break anything

cli/spark/aztklib.py
```diff
@@ -7,21 +7,4 @@
...
-            storage_account_suffix=secrets_config.storage_account_suffix,
-            docker_endpoint=secrets_config.docker_endpoint,
-            docker_password=secrets_config.docker_password,
-            docker_username=secrets_config.docker_username,
-            ssh_pub_key=secrets_config.ssh_pub_key,
-            ssh_priv_key=secrets_config.ssh_priv_key
-        )
-    )
\ No newline at end of file
+    return aztk.spark.Client(config.load_aztk_screts())
```
jafreck: This file seems less necessary now.

aztk/models.py
```diff
@@ -47,47 +50,137 @@ def __init__(self, ip_address, port):
...
+            self.service_principal.validate()
+        elif self.shared_key:
+            self.shared_key.validate()
+        else:
+            raise error.AztkError(
+                "Neither service_principal and shared_key auth are configured, must use only one")
+
+    def is_aad(self):
+        return self.service_principal is not None
+
```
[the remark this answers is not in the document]
timotheeguerin: actually the formatter is putting 2 spaces between classes

cli/spark/aztklib.py
timotheeguerin: yeah

cli/config.py
```diff
@@ -3,121 +3,115 @@
...
-        # read current working directory secrets.yaml
-        self._load_secrets_config()
+    default_config = secrets_config.get('default')
+    # Check for ssh keys if they are provided
+    if default_config:
+        secrets.ssh_priv_key = default_config.get('ssh_priv_key')
+        secrets.ssh_pub_key = default_config.get('ssh_pub_key')
 
 
 class ClusterConfig:
```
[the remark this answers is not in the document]
timotheeguerin: yeah could be a good idea, I can take a look at that in another pr

aztk/models.py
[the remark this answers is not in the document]
jafreck: yeah but aren't there 3 new lines here?

cli/spark/aztklib.py
jafreck: I think it could go directly in each of the endpoints. Like here: https://github.com/Azure/aztk/blob/1693f6ef20574afece0ce13f32c0a29697e5bd61/cli/spark/endpoints/cluster_add_user.py#L24

Just replace with `spark_client = aztk.spark.Client(config.load_aztk_screts())`

timotheeguerin: yeah actually could do that, i'll change it

cli/spark/endpoints/cluster_app_logs.py
```diff
@@ -1,8 +1,9 @@
 import argparse
 import typing
 import time
-from cli.spark.aztklib import load_spark_client
-from cli import utils
+import aztk
+import aztk
```
jafreck: duplicate

config/secrets.yaml.template
```diff
@@ -2,20 +2,20 @@
...
+    storage_account_resource_id:
 
-sharedKey:
-    # batchaccountname: 
-    # batchaccountkey: 
-    # batchserviceurl: 
-    # storageaccountname: 
-    # storageaccountkey: 
-    # storageaccountsuffix: core.windows.net
+shared_key:
```
emlyn: I think this should be commented out too by default, otherwise the verification will fail.

docs/50-sdk.md
```diff
@@ -388,22 +388,44 @@ Find some samples and getting stated tutorial in the `examples/sdk/` directory o
     The Batch, Storage, Docker and SSH secrets used to create AZTK clusters. For more help with setting these values see [Getting Started](./00-getting-started.md).
 
     Fields:
+        service_principal: ServicePrincipalConfiguration
+        shared_key: SharedKeyConfiguration
+        docker: DockerConfiguration
 
+        ssh_pub_key: str
+        ssh_priv_key: str
```
emlyn: Might be worth mentioning that exactly one of `service_principal` and `shared_key` must be provided.

aztk/utils/azure_api.py
```diff
@@ -16,80 +16,47 @@
...
         # Set up ServicePrincipalCredentials
-        credentials = ServicePrincipalCredentials(
-            client_id=batch_config.client_id,
-            secret=batch_config.credential,
-            tenant=batch_config.tenant_id,
+        arm_credentials = ServicePrincipalCredentials(
+            client_id=secrets.service_principal.client_id,
+            secret=secrets.service_principal.credential,
+            tenant=secrets.service_principal.tenant_id,
             resource='https://management.core.windows.net/')
```
emlyn: Is it OK to hard code this url (and the batch one below)? Are there any situations where a user might need to change them (e.g. maybe China)?

examples/sdk/sdk_example.py
```diff
@@ -6,11 +6,11 @@
 # set your secrets
 secrets_confg = aztk.spark.models.SecretsConfiguration(
     service_principal=aztk.models.ServicePrincipalConfiguration(
-        tenant_id = "<org>.onmicrosoft.com",
-        client_id = "",
-        credential = "",
-        batch_account_resource_id = "",
-        storage_account_resource_id = "",
+        tenant_id="<org>.onmicrosoft.com",
```
emlyn: Is this what a tenant_id typically looks like? Mine look like UUIDs.

timotheeguerin: yeah you can use the uuid or also the name, I tested that last week and it seems to work. This just look more friendly and its easier to find on the portal.

emlyn:
@timotheeguerin my authentication errors were a problem on my end as suspected, so I'm happy with this.

PR #343, opened 2018-01-22
Feature: Refactor cluster config to use ClusterConfiguration model

timotheeguerin:


aztk/spark/client.py
```diff
@@ -20,6 +20,12 @@ def __init__(self, secrets_config):
     Spark client public interface
     '''
     def create_cluster(self, cluster_conf: models.ClusterConfiguration, wait: bool = False):
+        cluster_conf.validate()
+
+        if cluster_conf.user_configuration.username is not None and wait is False:
+            raise error.AztkError(
```
paselem: I think this is going away soon w/ one of Jake's PRs

timotheeguerin: yeah with the create user on the node right? I just moved the validation here as it is still valid until the other PR get merged

cli/config.py
```diff
@@ -114,119 +114,64 @@ def _merge_secrets_dict(secrets: SecretsConfiguration, secrets_config):
...
-                "Please supply a vm_size in either the cluster.yaml configuration file or with a parameter (--vm-size)")
-
-        if self.wait is None:
-            raise aztk.error.AztkError(
-                "Please supply a value for wait in either the cluster.yaml configuration file or with a parameter (--wait or --no-wait)")
-
-        if self.username is not None and self.wait is False:
-            raise aztk.error.AztkError(
-                "You cannot create a user '{0}' if wait is set to false. By default, we create a user in the cluster.yaml file. Please either the configure your cluster.yaml file or set the parameter (--wait)".format(self.username))
+    return output, wait
```
paselem: seems weird to me to return two values here.. they don't really seem to go together.

jafreck: I believe he did it this way because cluster.yaml has a "wait" parameter, while the ClusterConfiguration object does not. 

The "wait" parameter is passed to client.create_cluster() along with the ClusterConfiguration so these are returned together here.

timotheeguerin: I could also return  `ClusterCLIParams` class here instead of a tuple

jafreck: I feel like a tuple works fine since the CLI is the only caller of this method.

aztk/spark/client.py
jafreck: I will have to update that PR to account for these changes.

aztk/models.py
```diff
@@ -23,26 +50,35 @@ def __init__(self, name: str = None, script: str = None, run_on=None):
         self.run_on = run_on
 
 
-class UserConfiguration:
+class UserConfiguration(ConfigurationBase):
```
jafreck: This should have a subclass in the `aztk.spark.models` package so people don't have to import `aztk.models` and `aztk.spark.models`.

cli/config.py
```diff
@@ -3,7 +3,7 @@
 import typing
 from cli import log
 import aztk.spark
-from aztk.models import ServicePrincipalConfiguration, SharedKeyConfiguration, DockerConfiguration
+from aztk.models import ServicePrincipalConfiguration, SharedKeyConfiguration, DockerConfiguration, ClusterConfiguration, UserConfiguration
```
jafreck: All of these should have subclasses in `aztk.spark.models`

cli/spark/endpoints/cluster/cluster_create.py
```diff
@@ -2,8 +2,10 @@
 import argparse
 import typing
 import aztk.spark
+from aztk.models import UserConfiguration
```
jafreck: same as above, subclass in `aztk.spark.models`

PR #344, opened 2018-01-22
Feature: Basic Cluster and Job Submission SDK Tests

### LINKED ISSUE #315: Add test suite
Tests should cover both the CLI and SDK. Manual integration tests will be needed to ensure functionality.

jafreck:
Fix #315


tests/spark/sdk/cluster/test_cluster.py
```diff
@@ -0,0 +1,143 @@
...
+    assert cluster.gpu_enabled is False
+    assert cluster.master_node_id is not None
+    assert cluster.current_low_pri_nodes == 0
+
+def test_list_clusters():
+    try:
+        clusters = spark_client.list_clusters()
+    except (AztkError, BatchErrorException) as e:
+        assert False
+    assert cluster_id in [cluster.id for cluster in clusters]
```
paselem: should we be testing more in this method? Like the number of clusters we expect to see?

tests/spark/sdk/job/test_job.py
```diff
@@ -0,0 +1,100 @@
...
+
+def test_get_application():
+    try:
+        application = spark_client.get_application(job_id=job_id, application_name="pipy100")
+    except (AztkError, BatchErrorException) as e:
+        assert False
+
+    assert isinstance(application, aztk.spark.models.Application)
+    assert application.exit_code == 0
+    assert application.state == "completed"
```
paselem: this check seems like it may be prone to failure based on when this method is run in relation to when the application was created.

tests/spark/sdk/cluster/test_cluster.py
jafreck: Since this can execute in any batch account, we couldn't be sure how many clusters there should be. Even if we get the count of clusters before we create one as part of these tests, we can't be sure that it will remain.

tests/spark/sdk/job/test_job.py
jafreck: pytest will execute tests from top of the file to the bottom. So by the time this runs, `test_submit_job` will have completed. And by the time that test has completed, the AZTK Job has completed, so the applications would have to also be in a "completed" state with exit code 0 for this test to succeed.

PR #358, opened 2018-01-25
Feature: spark init docker repo customization

### LINKED ISSUE #356: tailor init to user preferences
Add sub commands to init:
- `aztk spark init python`
- `aztk spark init r-base`
- `aztk spark init java/scala`

These specialized inits should set sensible defaults for docker_repo (eventually they could even come with suggested plugins).

jafreck:
Fix #356


cli/spark/endpoints/init.py
```diff
@@ -8,14 +8,29 @@
 def setup_parser(parser: argparse.ArgumentParser):
     parser.add_argument('--global', dest='global_flag', action='store_true',
                         help="Create a .aztk/ folder in your home directory for global configurations.")
+    software_parser = parser.add_mutually_exclusive_group()
+    software_parser.add_argument('--python', action="store_true", required=False)
+    software_parser.add_argument('--r-base', action="store_true", required=False)
```
paselem: Maybe just call it 'r' instead of 'r-base'? My guess is that these should be customer/solution focused, not based on our docker images.

cli/spark/endpoints/init.py
```diff
@@ -8,14 +8,29 @@
...
+    software_parser.add_argument('--java', action="store_true", required=False)
+    software_parser.add_argument('--scala', action="store_true", required=False)
```
paselem: Will it be confusing to users to split out java & scala? They're essentially the same thing and may leave customers wondering why we have split them out. Thoughts?

jafreck: No preference, just not sure how to word the flag if we combine them.

Under the hood, `--java`, `--scala` and no flag all do the same thing.

paselem: yeah.. I just can't think of a word that combines them both, like .NET but for java... I was thinking 'default' but that doesn't seem right either.

paselem: maybe just 'base'? and add some verbosity to the CLI and docs that this will use the out-of-the-box java/scale environment?

cli/spark/endpoints/init.py
```diff
@@ -8,14 +8,29 @@
 def setup_parser(parser: argparse.ArgumentParser):
     parser.add_argument('--global', dest='global_flag', action='store_true',
                         help="Create a .aztk/ folder in your home directory for global configurations.")
+    software_parser = parser.add_mutually_exclusive_group()
+    software_parser.add_argument('--python', action="store_true", required=False)
+    software_parser.add_argument('--r', action="store_true", required=False)
```
paselem: Can we accept 'r' or 'R'? I think the product is officially upper-case R.

cli/spark/endpoints/init.py
```diff
@@ -8,14 +8,29 @@
...
+    software_parser.add_argument('--java', action="store_true", required=False)
+    software_parser.add_argument('--scala', action="store_true", required=False)
+
 
 def execute(args: typing.NamedTuple):
+    # software_specific init
+    if args.python:
```
paselem: minor: good candidate for a switch statement. No need to change.

docs/00-getting-started.md
```diff
@@ -30,6 +30,14 @@ The minimum requirements to get started with this package are:
     ```
     This will create a *.aztk* folder with preset configuration files in your current working directory.
 
+    If you would like to initialize your AZTK clusters with a specific development toolset, please pass one of the following flags:
+    ```bash
+    aztk spark init --python
+    aztk spark init --r-base
```
paselem: typo? Should this be
```sh
aztk spark init --r?
```

cli/spark/endpoints/init.py
```diff
@@ -8,14 +8,29 @@
...
+    software_parser.add_argument('--r', '--R', action="store_true", required=False)
+    software_parser.add_argument('--java', action="store_true", required=False)
+    software_parser.add_argument('--scala', action="store_true", required=False)
+
 
 def execute(args: typing.NamedTuple):
+    # software_specific init
+    if args.python:
+        docker_repo = constants.DEFAULT_SPARK_PYTHON_DOCKER_REPO
+    elif args.r:
```
paselem: does this need to change to args.r || args.R?

jafreck: Both `--r` and `--R` go to `args.r`

PR #365, opened 2018-02-02
Allow submitting jobs into a VNET

emlyn:
I added a `subnet_id` option to the cluster config to allow creating the cluster in a VNET (as is possible in `aztk spark cluster create`)


jafreck:
This looks great, just needs some docs. I created a PR on your fork to add some.

emlyn:
@jafreck thanks, I've merged your PR.

PR #367, opened 2018-02-03
Bug: Load default Jars for job submission CLI

### LINKED ISSUE #366: Cannot access wasbs filesystem from submitted job
When I run an application via job submission, it looks like the azure storage jars are not added to the classpath (I see `java.io.IOException: No FileSystem for scheme: wasbs`). I've tried adding them to the application's `jars:` key in job.yaml, but this doesn't help.
In the log output I also see:
```
Warning: Local jar /home/spark-current/jars/azure-storage-2.0.0.jar does not exist, skipping.
Warning: Local jar /home/spark-current/jars/hadoop-azure-2.7.3.jar does not exist, skipping.
Warning: Local jar /home/spark-current/jars/azure-data-lake-store-sdk-2.2.3.jar does not exist, skipping.
```
Followed by:
```
java.io.FileNotFoundException: Jar /home/spark-current/jars/azure-storage-2.0.0.jar not found
	at org.apache.spark.SparkContext.addJarFile$1(SparkContext.scala:1807)
	at org.apache.spark.SparkContext.addJar(SparkContext.scala:1833)
	at org.apache.spark.SparkContext$$anonfun$12.apply(SparkContext.scala:466)
	at org.apache.spark.SparkContext$$anonfun$12.apply(SparkContext.scala:466)
	at scala.collection.immutable.List.foreach(List.scala:381)
	at org.apache.spark.SparkContext.<init>(SparkContext.scala:466)
	at org.apache.spark.SparkContext$.getOrCreate(SparkContext.scala:2509)
	at org.apache.spark.sql.SparkSession$Builder$$anonfun$6.apply(SparkSession.scala:909)
	at org.apache.spark.sql.SparkSession$Builder$$anonfun$6.apply(SparkSession.scala:901)
	at scala.Option.getOrElse(Option.scala:121)
	at org.apache.spark.sql.SparkSession$Builder.getOrCreate(SparkSession.scala:901)
```
And similar for `hadoop-azure-2.7.3.jar`, `azure-data-lake-store-sdk-2.2.3.jar`.
Do I need to do something in the config to make this work, or is it a bug?

jafreck:
Fix #366

### DIFF
--- a/cli/config.py
+++ b/cli/config.py
@@ -399,33 +399,24 @@ class JobConfig():
                     "No path to application specified for {} in job.yaml".format(entry['name']))
 
 
-def get_file_if_exists(file, local: bool):
-    if local:
-        if os.path.exists(os.path.join(aztk.utils.constants.DEFAULT_SPARK_CONF_SOURCE, file)):
-            return os.path.join(aztk.utils.constants.DEFAULT_SPARK_CONF_SOURCE, file)
-    else:
-        if os.path.exists(os.path.join(aztk.utils.constants.GLOBAL_CONFIG_PATH, file)):
-            return os.path.join(aztk.utils.constants.GLOBAL_CONFIG_PATH, file)
+def get_file_if_exists(file):
+    local_conf_file = os.path.join(aztk.utils.constants.DEFAULT_SPARK_CONF_SOURCE, file)
+    global_conf_file = os.path.join(aztk.utils.constants.GLOBAL_CONFIG_PATH, file)
+
+    if os.path.exists(local_conf_file):
+        return local_conf_file
+    if os.path.exists(global_conf_file):
+        return global_conf_file
+
+    return None
 
 
 def load_aztk_spark_config():
-    spark_defaults_conf = spark_env_sh = core_site_xml = None
-
-    spark_defaults_conf = get_file_if_exists('spark-defaults.conf', False)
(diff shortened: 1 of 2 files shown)


cli/config.py
```diff
@@ -398,16 +399,37 @@ def merge(self, id, job_config_yaml=None):
...
-                return os.path.join(aztk.utils.constants.DEFAULT_SPARK_CONF_SOURCE, file)
-        else:
-            if os.path.exists(os.path.join(aztk.utils.constants.GLOBAL_CONFIG_PATH, file)):
-                return os.path.join(aztk.utils.constants.GLOBAL_CONFIG_PATH, file)
+    spark_defaults_conf = spark_env_sh = core_site_xml = None
+
+    spark_defaults_conf = get_file_if_exists('spark-defaults.conf', False)
+    spark_env_sh = get_file_if_exists('spark-env.sh', False)
+    core_site_xml = get_file_if_exists('core-site.xml', False)
+
```
paselem: I'm not sure I understand why you're making the same 3 calls with False, then True - seems kind of strange to me. Is this mean to override the defaults?

jafreck: Agreed this is poorly written, but what it's doing is getting the global configuration file, and then overwriting it with the local configuration file if it exists. I will rewrite this before I merge -- too clunky as is.

PR #374, opened 2018-02-07
Feature: spark shuffle service

### LINKED ISSUE #370: Add Spark Shuffle Service Support
Enable shuffle service by default by setting `spark.shuffle.service.enabled true` in spark-defaults.conf.

Also, open port 7337 in container.

https://spark.apache.org/docs/latest/configuration.html#shuffle-behavior

jafreck:
Fix #370


config/spark-defaults.conf
```diff
@@ -31,3 +31,5 @@ spark.jars                      /home/spark-current/jars/azure-storage-2.0.0.jar
 # Note: Default filesystem master HA
 spark.deploy.recoveryMode       FILESYSTEM
 spark.deploy.recoveryDirectory  /root/
+
+spark.shuffle.service.enabled   true
```
paselem: as part of this change, should we also move the default directory off of the OS drive? It feels a bit worrying to me that we're intentionally going to put data there.

node_scripts/install/spark.py
```diff
@@ -114,6 +114,15 @@ def start_spark_worker():
...
     call(cmd)
 
+    # enable the shuffle service on all slaves
+    start_shuffle_service()
+
+
+def start_shuffle_service():
+    exe = os.path.join(spark_home, "sbin", "start-shuffle-service.sh")
+    print("Starting the shuffle service with {}".format(exe))
+    call([exe, " &"])
```
paselem: What happens if this fails to start?

jafreck: If the shuffle service fails to start, it just won't be used.

PR #388, opened 2018-02-09
Fix: Custom scripts not read from cluster.yaml

timotheeguerin:


cli/config.py
```diff
@@ -160,11 +161,27 @@ def cluster_config_from_dict(config: dict):
         if config.get('password') is not None:
             output.user_configuration.password = config['password']
 
-    if config.get('custom_scripts') not in [[None], None]:
-        output.custom_scripts = config['custom_scripts']
+    if 'custom_scripts' in config:
```
jafreck: Why no `if config.get('custom_scripts) is not None` like the above?

cli/config.py
```diff
@@ -160,11 +161,27 @@ def cluster_config_from_dict(config: dict):
...
+            output.custom_scripts.append(
+                aztk.spark.models.CustomScript(
+                    script=custom_script['script'],
+                    run_on=custom_script['runOn']
+                )
+            )
 
-    if config.get('azure_files') not in [[None], None]:
-        output.file_shares = config['azure_files']
+    if 'azure_files' in config:
```
jafreck: same here

cli/config.py
```diff
@@ -305,16 +323,21 @@ def _merge_dict(self, config):
 
         spark_configuration = config.get('spark_configuration')
         if spark_configuration:
-            self.spark_defaults_conf = self.__convert_to_path(spark_configuration.get('spark_defaults_conf'))
-            self.spark_env_sh = self.__convert_to_path(spark_configuration.get('spark_env_sh'))
-            self.core_site_xml = self.__convert_to_path(spark_configuration.get('core_site_xml'))
-            self.jars = [self.__convert_to_path(jar) for jar in spark_configuration.get('jars')]
+            self.spark_defaults_conf = self.__convert_to_path(
+                spark_configuration.get('spark_defaults_conf'))
```
jafreck: I feel like splits like this make the code harder to read

cli/config.py
timotheeguerin: Yeah, felt like it was looking cleaner hut this doesn't cover the null case

cli/config.py
timotheeguerin: The auto formatter I think format at x number of column

cli/config.py
timotheeguerin: though why was it checking for [[None], None]?

cli/config.py
timotheeguerin: I think we need to look into the formatter so we all have the same settings, max column length, etc. So the editor doesn't format

timotheeguerin: ok so I added yapf and changed vscode default to use that for formatting, it seems like autopep8(default) doesn't listen to the max line length rule. I set the max line length to 120(Instead of default 80) when yapf format it reforms a lot of the splitted lines which makes a lot of changes for this PR so I won't format this file this time(just revert those changes).

PR #392, opened 2018-02-09
Bug: stop using mutable default parameters

### LINKED ISSUE #389: Stop using mutable default parameters
A couple functions have mutable default arguments. This is not safe as python evaluates the arguments once, not each time the function is called. 

Described more here: http://docs.python-guide.org/en/latest/writing/gotchas/#mutable-default-arguments

jafreck:
Fix #389

### DIFF
--- a/aztk/spark/models.py
+++ b/aztk/spark/models.py
@@ -123,9 +123,9 @@ class ApplicationConfiguration:
             application=None,
             application_args=None,
             main_class=None,
-            jars=[],
-            py_files=[],
-            files=[],
+            jars=None,
+            py_files=None,
+            files=None,
             driver_java_options=None,
             driver_library_path=None,
             driver_class_path=None,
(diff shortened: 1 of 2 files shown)


aztk/spark/models.py
```diff
@@ -123,9 +123,9 @@ def __init__(
             application=None,
             application_args=None,
             main_class=None,
-            jars=[],
-            py_files=[],
-            files=[],
+            jars=None,
```
timotheeguerin: don't you need to do something like `self.jars = jars or []` instead to keep the same

jafreck: That's what line [141](https://github.com/Azure/aztk/pull/392/files#diff-73e8698c9fc8e05c277a1850e41dac89R141) does unless I'm misunderstanding what you're saying.

timotheeguerin: oh nvm missed that, didn't saw it for the ones above but it actually check for not null so it doesn't matter

PR #395, opened 2018-02-12
Bug: always upload spark job logs, errors

### LINKED ISSUE #393: job submission app not uploading logs
I'm submitting a job with `aztk spark job submit`. When it completes I try to get the logs with `aztk spark job get-app-logs --id xxx --name yyy`, but I just get an error saying:
```
Logs not found in your storage account. They were either deleted or never existed.
```
Indeed the logs don't appear to be in the storage account, which makes it quite hard to debug failures. I'm pretty sure this used to work when I first tried using job submission.

Let me know if you need any extra information to help reproduce it.

jafreck:
Fix #393

### DIFF
--- a/node_scripts/submit.py
+++ b/node_scripts/submit.py
@@ -162,18 +162,32 @@ def recieve_submit_request(application_file_path):
         driver_cores=application['driver_cores'],
         executor_cores=application['executor_cores'])
 
-    try:
-        return_code = subprocess.call(cmd.to_str(), shell=True)
-        upload_log(blob_client, application)
-        return return_code
-    except subprocess.CalledProcessError as e:
-        print(e)
-        upload_log(blob_client, application)
-        return 1
+    return_code = subprocess.call(cmd.to_str(), shell=True)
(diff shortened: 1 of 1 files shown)


node_scripts/submit.py
```diff
@@ -162,18 +162,32 @@ def recieve_submit_request(application_file_path):
...
-    except subprocess.CalledProcessError as e:
-        print(e)
-        upload_log(blob_client, application)
-        return 1
+    return_code = subprocess.call(cmd.to_str(), shell=True)
+    upload_log(blob_client, application)
+    return return_code
+
+
+def upload_error_log(error, application_file_path):
```
paselem: Should we _always_ upload these? Maybe an option to override at job submit?

jafreck: I think we should always upload on an error. If you don't there is no explicit way to tell that something went wrong, or what went wrong. 

On a successful run, error logs won't be uploaded. Note that a successful run includes when a spark application has failed. An error here is only when something goes wrong on AZTK's end.

PR #401, opened 2018-02-13
Internal: Cluster data helpers and upload_node_script into cluster_data module

timotheeguerin:
helpers and upload_node_script was getting quite messy.
Created a new module `cluster_data` which goal is to manage all the cluster data(Blob container).


aztk/client.py
```diff
@@ -27,6 +27,13 @@ def __init__(self, secrets_config: models.SecretsConfiguration):
         self.blob_client = azure_api.make_blob_client(secrets_config)
 
 
+
+    def get_cluster_data(self, cluster_id: str) -> cluster_data.ClusterData:
```
jafreck: This is meant to be for internal use only, right? So maybe make the function `_get_cluster_data` instead.

aztk/internal/cluster_data/blob_data.py
```diff
@@ -0,0 +1,26 @@
...
+    Object mapping to a blob entry. Can generate resource files for batch
+    """
+    def __init__(self, blob_client: BlockBlobService, container: str, blob: str):
+        self.container = container
+        self.blob = blob
+        self.dest = blob
+        self.blob_client = blob_client
+
+
+    def as_resource_file(self, dest: str = None) -> batch_models.ResourceFile:
```
jafreck: I feel like this should be `to_resource_file`.

aztk/internal/cluster_data/cluster_data.py
```diff
@@ -0,0 +1,57 @@
...
+        except azure.common.AzureMissingResourceHttpError:
+            logging.warn("Cluster %s doesn't have cluster configuration in storage", self.cluster_id)
+        except yaml.YAMLError:
+            logging.warn("Cluster %s contains invalid cluster configuration in blob", self.cluster_id)
+
+    def upload_file(self, blob_path: str, local_path: str) -> BlobData:
+        self.blob_client.create_blob_from_path(self.cluster_id, blob_path, local_path)
+        return BlobData(self.blob_client, self.cluster_id, blob_path)
+
+    def upload_cluster_file(self, blob_path: str, local_path: str) -> BlobData:
```
jafreck: Does this need to exist? Looks like it's only called by `upload_node_data` below, they could probably be combined. Or are we anticipating that files other than node_scripts will be put in the `cluster` dir?

aztk/internal/cluster_data/node_data.py
```diff
@@ -0,0 +1,166 @@
...
+    Class made to bundle data to be uploaded to the node as a zip
+    """
+
+    def __init__(self, cluster_config: models.ClusterConfiguration):
+        self.zip_path = os.path.join(ROOT_PATH, "tmp/node-scripts.zip")
+        self.cluster_config = cluster_config
+        file_utils.ensure_dir(self.zip_path)
+        self.zipf = zipfile.ZipFile(self.zip_path, "w", zipfile.ZIP_DEFLATED)
+
+    def __enter__(self):
```
jafreck: This is nice, and we might want to add it a few other places, but it doesn't look like you used it at all for NodeData?

aztk/internal/cluster_data/cluster_data.py
timotheeguerin: yeah was having this one in case there is any other files that would need to be uploaded. Feel like it doesn't hurt to have it here

aztk/internal/cluster_data/node_data.py
timotheeguerin: yeah I actually didn't ended up using for node data so could remove it here but there is a few place where this could be useful

PR #413, opened 2018-02-22
Feature: task affinity to master node

### LINKED ISSUE #412: Spark drvier should run on master for single app job submission
Scoped version of #375 for single application job submission mode

jafreck:
Fix #412

### DIFF
--- a/node_scripts/job_submission.py
+++ b/node_scripts/job_submission.py
@@ -1,13 +1,22 @@
-import sys
-import os
-import yaml
-import subprocess
 import datetime
+import os
+import subprocess
+import sys
 from typing import List
-import azure.storage.blob as blob
 import azure.batch.models as batch_models
+import azure.storage.blob as blob
+import yaml
 from command_builder import CommandBuilder
 from core import config
+from install.pick_master import get_master_node_id
+
+
+def affinitize_task_to_master(batch_client, cluster_id, task):
+    pool = batch_client.pool.get(config.pool_id)
+    master_node_id = get_master_node_id(pool)
+    master_node = batch_client.compute_node.get(pool_id=cluster_id, node_id=master_node_id)
+    task.affinity_info = batch_models.AffinityInformation(affinity_id=master_node.affinity_id)
+    return task
 
 
 def schedule_tasks(tasks_path):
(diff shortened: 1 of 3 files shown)


node_scripts/job_submission.py
```diff
@@ -1,13 +1,22 @@
...
-import azure.storage.blob as blob
 import azure.batch.models as batch_models
+import azure.storage.blob as blob
+import yaml
 from command_builder import CommandBuilder
 from core import config
+from install.pick_master import get_master_node_id
+
+
+def affinitize_task_to_master(batch_client, cluster_id, task):
```
paselem: I know we've talked about this before, but it is still strange to me that this code is essentially duplicated.

node_scripts/submit.py
```diff
@@ -105,8 +105,10 @@ def __app_submit_cmd(
     spark_submit_cmd.add_option('--driver-class-path', driver_class_path)
     spark_submit_cmd.add_option('--driver-memory', driver_memory)
     spark_submit_cmd.add_option('--executor-memory', executor_memory)
-    spark_submit_cmd.add_option('--driver-cores', str(driver_cores))
-    spark_submit_cmd.add_option('--executor-cores', str(executor_cores))
+    if driver_cores:
```
paselem: shouldn't we do this for all numerical settings (e.g. executor_memory)?

node_scripts/job_submission.py
jafreck: yeah -- definitely not ideal, but this is one of the only options until we do the larger refactor we've discussed around loading the entire SDK (not just node scripts) as a resource file for the cluster. That refactor should simplify/eliminate a lot of the duplicate code we have in node scripts.

node_scripts/submit.py
jafreck: I believe, executor memory is already a string (a numerical value and a unit) like "4g" or "1024M"

paselem:
Approved but have one final comment... what happens if the task doesn't affinitize to the master? (I know what actually happens, but what does the customer see? Will it affect them at all? Should we block?)

jafreck:
The customer won't see anything if the task does not affinitize to the master. It could potential affect them if they expect the driver to execute on the master node only, but we don't claim that the driver always executes on the node (yet, at least). This was just meant to be a patch to unblock people who submit 1 application Jobs.

A real fix for establishing a known location for the driver is still in the works.

paselem:
Ok, that makes sense. I think that it's good that we don't advertise it since we don't guarantee it right now.

PR #428, opened 2018-03-01
Feature: Pypi auto deployement 

timotheeguerin:
Todo before merge:
- [x] Revert `feature/pypi` specific
- [x] Setup build on tag only


.travis.yml
```diff
@@ -13,4 +13,14 @@ script:
...
+    - master
+
+deploy:
+  provider: pypi
+  distributions: "sdist bdist_wheel"
+  user: aztk
+  password:
+    secure: j/mwA+hWudujDZ+JkgN3hDBUcIH3Vt1SWsvxnyUadErFsi9S9bW8OZ8kOp3R8yj73pnaENADmvDPjBt6w39b3h7l/EmNYlEdsAuPBMmaNQ0+fmPiKLCcALVIt/Odxdc+a1p2p+F3HEatqMCkaSRUs9gcHSYA1P57rt7y6i28xrCTt1ayQrdOZofgyAx4egocFQlgXgFpQIO3vNXySmc5HaIJF8h84CUKvLG/jv9bHgDoT7n2F00sQQV/tLbXdWEA2LcJFN5q4gjk3AwFtofTtMzlk3IE3y3qp9DagWbdsKoy9+b31lypVDBiIp/N+kCm/pwYDJXyG0H7U1pev2F8K/f4xF2x0AJDJJamLczAY4Ac4AMPImlI3cCpSS6htQhb+i5+KeEUW2tm+xGmdprOSto2712C4TNMna25x+WZteuPmRGYTmmDxGSeD769lE6TMGqrJTlmpkzndJiyp8ek/86hJPJUQoPfDadWdIdcA4wCjpXFBach4DYEItKvYoIk/KfK3wdqp3Lbs+MhT+JhuXGR0RRWlkCROiIo48TwQ716ddvjy2PPxwxNhplu3BdyEXLEDGN9EnEL+3L5P0GELhB4+KstjXSbECBYzULJ1+4kzvrdSB3K9F41stpT7x9PN9mK0t9FrVpXWbNozuJqkCvVVtz2ZSJ+20kvSuI1TPA=
+  on:
+    python: 3.6
```
jafreck: I'm not exactly sure here, but does this not need to also have `python: 3.5`?

setup.py
```diff
@@ -1,22 +1,62 @@
...
+
+setup(
+    name='aztk',
+    version=version.__version__,
+    description=
+    'Utility for data engineers or platform developers to Run distributed jobs in Azure',
+    long_description=long_description,
+    url='https://github.com/Azure/aztk',
+    author='Microsoft',
+    author_email='jiata@microsoft.com',
```
jafreck: Should be switched to `askaztk@microsoft.com` I think.

setup.py
```diff
@@ -1,22 +1,62 @@
...
+
+
+with open('README.md') as fd:
+    long_description = fd.read()
+
+setup(
+    name='aztk',
+    version=version.__version__,
+    description=
+    'Utility for data engineers or platform developers to Run distributed jobs in Azure',
```
jafreck: Shouldn't this match the repo's tagline? "On-demand, Dockerized, Spark Jobs on Azure (powered by Azure Batch)"

.travis.yml
timotheeguerin: no this means I only want to publish the build of version 3.6(could be 3.5 too) just picked the latest

setup.py
timotheeguerin: yeah didn't touch that one, could change it

setup.py
jafreck: This still needs to be updated I think.

PR #435, opened 2018-03-13
Fix job configuration option for `aztk spark job submit` command

shtratos:
`--job-conf` option mentioned in the docs wasn't working.

CLI help was showing that option is named `--configuration-c`
which seems to be a result of a missing comma in option definition.


jafreck:
Fix #436

PR #437, opened 2018-03-13
Enable pulling docker images from private docker registry

stevekuo4:
There is a typo in setup_node.sh that prevents one from pulling image from a private docker registry.
Also when pulling image, there is no need to append DOCKER_ENDPOINT to repo_name.


seddonm1:
thanks @stevekuo4. just spent hours debugging and was going to submit a pull request before i found yours. 👍

PR #443, opened 2018-03-16
Feature: managed storage for clusters and jobs

### LINKED ISSUE #419: Manage storage account
Storage account should be managed.

- [x] Delete cluster blob container on `cluster delete`

- [x] Delete job blob container on `job delete`

- [ ] Add `--keep-logs` flag that prevents storage deletion

jafreck:
Fix #419 

TODO:
- [x] Add `--keep-logs` flag to prevent storage deletion
- [x] Add docs for storage lifetime, and `--keep-logs` flag

### DIFF
--- a/aztk/client.py
+++ b/aztk/client.py
@@ -61,6 +61,9 @@ class Client:
 
         if pool_exists:
             self.batch_client.pool.delete(pool_id)
+        
+        cluster_data = self._get_cluster_data(pool_id)
+        cluster_data.delete_container(pool_id)
 
         return job_exists or pool_exists
 
--- a/aztk/client.py
+++ b/aztk/client.py
@@ -39,7 +39,7 @@ class Client:
     General Batch Operations
     '''
 
-    def __delete_pool_and_job(self, pool_id: str):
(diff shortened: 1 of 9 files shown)


aztk/client.py
```diff
@@ -61,6 +61,9 @@ def __delete_pool_and_job(self, pool_id: str):
 
         if pool_exists:
             self.batch_client.pool.delete(pool_id)
+        
+        cluster_data = self._get_cluster_data(pool_id)
+        cluster_data.delete_container(pool_id)
```
timotheeguerin: sweet!

aztk_cli/spark/endpoints/cluster/cluster_delete.py
```diff
@@ -22,6 +22,7 @@ def execute(args: typing.NamedTuple):
     cluster_id = args.cluster_id
 
     if not args.force:
+        log.warn("All logs persisted for this cluster will be deleted.")
```
timotheeguerin: hhm, that does seem like a good idea

aztk/client.py
jafreck: yeah this was a super simple change because of the `cluster_data` refactor

PR #455, opened 2018-03-22
Feature: spark debug tool

### LINKED ISSUE #347: Add a dump logs command for debugging user issues
This can be a flag for `aztk spark cluster run` or something like that. 

It should: 
1. ssh into each node (create 'aztk' user on the fly)
2. get stdout.txt
3. get the docker logs
4. executor logs
5. driver logs

jafreck:
Fix #347


aztk/utils/ssh.py
```diff
@@ -40,59 +40,84 @@ def connect(hostname,
...
     client = connect(hostname=hostname, port=port, username=username, password=password, pkey=ssh_key)
-    docker_exec = 'sudo docker exec 2>&1 -t {0} /bin/bash -c \'set -e; set -o pipefail; {1}; wait\''.format(container_name, command)
-    stdin, stdout, stderr = client.exec_command(docker_exec, get_pty=True)
-    [print(line.decode('utf-8')) for line in stdout.read().splitlines()]
+    if container_name:
+        cmd = 'sudo docker exec 2>&1 -t {0} /bin/bash -c \'set -e; set -o pipefail; {1}; wait\''.format(container_name, command)
+    else:
+        cmd = '/bin/bash 2>&1 -c \'set -e; set -o pipefail; {0}; wait\''.format(command)
+    stdin, stdout, stderr = client.exec_command(cmd, get_pty=True)
+    # [print(line.decode('utf-8')) for line in stdout.read().splitlines()]
```
paselem: remove commented log?

aztk_cli/spark/endpoints/cluster/cluster_run.py
```diff
@@ -13,5 +13,9 @@ def setup_parser(parser: argparse.ArgumentParser):
 
 def execute(args: typing.NamedTuple):
     spark_client = aztk.spark.Client(config.load_aztk_secrets())
-    result = spark_client.cluster_run(args.cluster_id, args.command)
+    results = spark_client.cluster_run(args.cluster_id, args.command)
+    for result in results:
+        print("---------------------------") #TODO: replace with nodename
```
paselem: todo?

aztk/spark/client.py
```diff
@@ -146,15 +147,16 @@ def get_application_status(self, cluster_id: str, app_name: str):
...
-    def cluster_run(self, cluster_id: str, command: str):
+    def cluster_run(self, cluster_id: str, command: str, host=False):
         try:
-            return self.__cluster_run(cluster_id, 'spark', command)
+            return self.__cluster_run(cluster_id, command, container_name='spark' if not host else None)
         except batch_error.BatchErrorException as e:
             raise error.AztkError(helpers.format_batch_exception(e))
 
-    def cluster_copy(self, cluster_id: str, source_path: str, destination_path: str):
+    def cluster_copy(self, cluster_id: str, source_path: str, destination_path: str, host=False, get=False):
```
timotheeguerin: what does `get` do?

aztk_cli/spark/endpoints/cluster/cluster_debug.py
```diff
@@ -0,0 +1,19 @@
...
+import typing
+import aztk.spark
+from aztk_cli import config
+
+
+def setup_parser(parser: argparse.ArgumentParser):
+    parser.add_argument('--id', dest='cluster_id', required=True,
+                        help='The unique id of your spark cluster')
+
+    parser.add_argument('--output', '-o', required=True,
```
timotheeguerin: couldn't we make that optional to be something like `aztk_debug/[cluster_id]` by default

pylintrc
```diff
@@ -9,7 +9,7 @@
 
 # Add files or directories to the blacklist. They should be base names, not
 # paths.
-ignore=CVS
+ignore=CVS,debug.py
```
timotheeguerin: why did you remove this one?

jafreck: I didn't remove it, I added debug.py to the pylint ignore list

aztk/spark/client.py
jafreck: get means retrieve files from the nodes. previously cluster copy was limited to copying a local file to all nodes on the cluster. Now it can work both ways.

aztk/spark/helpers/cluster_diagnostic_helper.py
```diff
@@ -0,0 +1,26 @@
...
+def run(spark_client, cluster_id, output_directory):
+    # copy debug program to each node
+    spark_client.cluster_copy(cluster_id, os.path.abspath("./aztk/spark/utils/debug.py"), "/tmp/debug.py", host=True)
+    ssh_cmd = _build_diagnostic_ssh_command()
+    run_output = spark_client.cluster_run(cluster_id, ssh_cmd, host=True)
+    local_path = os.path.join(os.path.abspath(output_directory), "debug", "debug.zip") #TODO: add timestamp
+    remote_path = "/tmp/debug.zip"
+    output = spark_client.cluster_copy(cluster_id, remote_path, local_path, host=True, get=True)
+    # write run output to debug/ directory
+    with open(os.path.join(os.path.dirname(local_path), "debug-output.txt"), 'w', encoding="UTF-8") as f:
```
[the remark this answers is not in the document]
jafreck: Not sure what you mean by this. debug-output.txt is the output from the cluster_run command. It is mainly just there to see if the tool crashed or not.
The file is also only written once so I'm not sure why I would overwrite.

aztk_cli/spark/endpoints/cluster/cluster_debug.py
jafreck: yeah, that's a good idea. by default, it can be `debug-{cluster-id}/` in the working directory.

aztk/spark/utils/debug.py
```diff
@@ -0,0 +1,160 @@
+"""
+    Diagnostic program that runs on each node in the cluster
+    This program must be run with sudo
+"""
+import io
```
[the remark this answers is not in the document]
jafreck: These are PEP8 sorted, and alpha sorted within the groups.

pylintrc
jafreck: No idea but it does not seem necessary at all.

jafreck: switched to single `# pylint: disable=import-error` on the import line and it seems to work fine.

aztk/spark/client.py
jafreck: what would you suggest? is it just that the name of the parameter confusing?

aztk/spark/helpers/cluster_diagnostic_helper.py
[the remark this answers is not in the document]
paselem: I was under the impression you could run the tool multiple times. If that happens, do we want to append or overwrite?

aztk/spark/utils/debug.py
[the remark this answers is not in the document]
paselem: I have no idea why I felt these were not sorted... You're right.

aztk/spark/client.py
paselem: Not sure if having two methods would be more clear?
```python
def copy_to_cluster(...):
    ...

def copy_from_cluster(...):
    ...
```

Thoguhts?

jafreck: A change like that would be breaking for any script using `cluster_copy()` today, unless we have `cluster_copy()` and `copy_from_cluster()`. I feel like that is not the best naming.

In general, I think we should consider an entire SDK rewrite to align cluster and job function names (didn't do a particularly good job naming them). It would also be nice to split the client so it has a `cluster` and a `job` submodule. So you would do `client.cluster.get_log()` or `client.job.get_log()`.

timotheeguerin: yeah I'm down to rename stuff. We can either keep the old stuff  to call the new function and mark as depractated or just remove them as we are technically only releasing it next version

timotheeguerin: i think deprecating is probably better with something like that https://stackoverflow.com/questions/2536307/how-do-i-deprecate-python-functions

jafreck: Depreciating is fine, I think so long as we have a set time frame (maybe 1 or 2 versions) where we actually remove the code.

PR #456, opened 2018-03-22
Bug: add spark.history.fs.logDirectory to required keys

### LINKED ISSUE #452: History server doesn't start
The spark history server doesn't appear to be running on the master instance at port 18080, as the splash screen suggests when a cluster is created.

I managed to get it to start on an already-running cluster by setting `spark.history.fs.logDirectory` to `/mnt/batch/spark-events` (event logs can get quite big, `/tmp` will likely fill up on a big job), and making the target directory before running the init script in `$SPARK_HOME/sbin`

Update: You actually also need to enable logging to the same location in the app itself to use the history server. The following configuration worked well for me:

```
spark.history.fs.logDirectory     /mnt/batch/tasks/spark-events
spark.eventLog.dir                /mnt/batch/tasks/spark-events
spark.eventLog.enabled            true
spark.eventLog.compress           true
```

jafreck:
Fix #452


node_scripts/install/spark.py
```diff
@@ -88,12 +88,14 @@ def start_history_server():
     # configure the history server
     spark_event_log_enabled_key = 'spark.eventLog.enabled'
     spark_event_log_directory_key = 'spark.eventLog.dir'
+    spark_history_fs_log_directory = 'spark.history.fs.logDirectory'
```
paselem: You should probably add this as a commented out field at https://github.com/Azure/aztk/blob/master/aztk_cli/config/spark-defaults.conf

node_scripts/install/spark.py
```diff
@@ -88,12 +88,14 @@ def start_history_server():
...
     path_to_spark_defaults_conf = os.path.join(spark_home, 'conf/spark-defaults.conf')
     properties = parse_configuration_file(path_to_spark_defaults_conf)
+    required_keys = [spark_event_log_directory_key, spark_history_fs_log_directory]
```
paselem: Missing spark_event_log_enabled?

I think this broke between Spark versions... Different Spark versions have different requirements, which makes this a difficult dictionary to maintain.

aztk/node_scripts/install/spark.py
```diff
@@ -207,6 +189,25 @@ def parse_configuration_file(path_to_file: str):
...
+    spark_history_fs_log_directory = 'spark.history.fs.logDirectory'
+    path_to_spark_defaults_conf = os.path.join(spark_home, 'conf/spark-defaults.conf')
+    properties = parse_configuration_file(path_to_spark_defaults_conf)
+    required_keys = [spark_event_log_enabled_key, spark_event_log_directory_key, spark_history_fs_log_directory]
+
+    # only enable the history server if it was enabled in the configuration file
+    if properties:
+        if all(key in properties for key in required_keys):
+            configure_history_server_log_path(properties[spark_history_fs_log_directory])
+        exe = os.path.join(spark_home, "sbin", "start-history-server.sh")
```
paselem: Shouldn't lines 205 - 208 be part of the if all(key in properties...) ? Otherwise it will always execute this even if the history server was not configured, right? Not sure if spark will just ignore this though since the properties will say it's a no-op, but still worth not even starting the process.

docs/13-configuration.md
```diff
@@ -1,113 +1,128 @@
...
+
+The core settings for a cluster are configured in the *cluster.yaml* file. Once you have set your desired values in *.aztk/cluster.yaml*, you can create a cluster using `aztk spark cluster create`.
+
+This is the default cluster configuration:
+
+```yaml
+# id: <id of the cluster to be created>
+id: spark_cluster
+
+# vm_size: <vm-size, see available options here: https://azure.microsoft.com/en-us/pricing/details/virtual-machines/linux/>
```
paselem: probably worth updating the link to batch pricing:
https://azure.microsoft.com/pricing/details/batch/

docs/13-configuration.md
```diff
@@ -1,113 +1,128 @@
...
+# job_ui_port: <local port where the job ui is forwarded to>
+job_ui_port: 4040
+
+# job_history_ui_port: <local port where the job history ui is forwarded to>
+job_history_ui_port: 18080
+
+# web_ui_port: <local port where the spark master web ui is forwarded to>
+web_ui_port: 8080
+
+# jupyter_port: <local port which where jupyter is forwarded to>
```
paselem: Should we get rid of the items below since they are now driven by plugins?

docs/13-configuration.md
```diff
@@ -1,113 +1,128 @@
...
+name_node_ui_port: 50070
+
+# rstudio_server_port: <local port which where rstudio server is forwarded to>
+rstudio_server_port: 8787
+
+# connect: <true/false, connect to spark master or print connection string (--no-connect)>
+connect: true
+```
+
+Running the command `aztk spark cluster ssh --id <cluster_id>` will ssh into the master node of the Spark cluster. It will also forward the Spark Job UI to localhost:4040, the Spark master's web UI to localhost:8080, and Jupyter to localhost:8888.
```
paselem: This is also out of date now.

docs/13-configuration.md
```diff
@@ -1,113 +1,128 @@
...
+
+Note that all of the settings in ssh.yaml will be overrided by parameters passed on the command line.
```
paselem: ... or the plugin

docs/13-configuration.md
```diff
@@ -1,113 +1,128 @@
...
+- Any options related to YARN client mode or Mesos
+
+`spark-defaults.conf`:
+- spark.master
+
+Also note that this toolkit pre-loads wasb jars, so loading them elsewhere is not necessary.
+
+### History Server
+If you want to use Spark's history server, please set the following values in your `.aztk/spark-defaults.conf` file:
+```
```
paselem: Worth linking off to the spark docs for this? (hard part would be to know which version of the docs though...)

docs/13-configuration.md
jafreck: yeah -- we should probably get rid of `ssh.yaml` entirely. The only thing it does now is set a default username. It's been almost completely replaced by plugins.

docs/13-configuration.md
jafreck: yeah --most of this file is out of date. All of the plugin docs update will come with #461 and #467

PR #459, opened 2018-03-26
Feature: JupyterLab plugin

paselem:
- Include Jupyter Lab as a shipped plugin
- Add some more verbosity to logs
- Write a MASTER sentinel file to disk to know where master is


node_scripts/install/install.py
```diff
@@ -27,6 +27,11 @@ def setup_node():
         plugins.setup_plugins(is_master=True, is_worker=True)
         scripts.run_custom_scripts(is_master=True, is_worker=True)
 
+        #Write sentinel file on disk for host to know this is the master
+        path = os.environ['PWD'] + '/MASTER'
+        print('Writing master file to: {}'.format(path))
+        f = open(path, 'w')
+        f.write(master_node.ip_address)
```
jafreck: This is already done, here:https://github.com/Azure/aztk/blob/c244e3437b74fa899e6b7450f642e3b165da0903/node_scripts/install/spark.py#L50

aztk/spark/models/plugins/jupyter_lab/jupyter_lab.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+
+# This custom script only works on images where jupyter is pre-installed on the Docker image
```
jafreck: We should change all of the plugins to be self-contained. Not sure that should be done in this PR, but we will have to change it soon anyways.

aztk/spark/models/plugins/jupyter_lab/jupyter_lab.sh
```diff
@@ -0,0 +1,57 @@
...
+    ],
+    "env": {
+        "SPARK_HOME": "$SPARK_HOME",
+        "PYSPARK_PYTHON": "python",
+        "PYSPARK_SUBMIT_ARGS": "--master spark://$MASTER_IP:7077 pyspark-shell"
+    }
+}
+EOF
+
+    # start jupyter notebook from /mnt - this is where we recommend you put your azure files mount point as well
```
jafreck: does jupyterlab have the same restriction as jupyter where you can't navigate up the directory structure? Right now `/mnt` is cluttered with a lot of things, it would be nice to have jupyterlab's default directory point somewhere else.

node_scripts/install/install.py
paselem: Removed

aztk/spark/models/plugins/jupyter_lab/jupyter_lab.sh
paselem: Agreed, although this is no longer true since I'm install jupyter lab from conda. I copied the file over so that comment was unnecessary.

aztk/spark/models/plugins/jupyter_lab/jupyter_lab.sh
paselem: Same limitation - you cannot navigate 'up' from /mnt. Since this is where we place samples it seems like a reasonable place to put it - where else do you think it could work?

jafreck: No that's fine, `/mnt/` is the best place since that is where their azure files will be mounted,

aztk/spark/models/plugins/jupyter_lab/configuration.py
```diff
@@ -0,0 +1,23 @@
...
+from aztk.models.plugins.plugin_configuration import PluginConfiguration, PluginPort, PluginRunTarget
+from aztk.models.plugins.plugin_file import PluginFile
+from aztk.utils import constants
+
+dir_path = os.path.dirname(os.path.realpath(__file__))
+
+class JupyterLabPlugin(PluginConfiguration):
+    def __init__(self):
+        super().__init__(
+            name="jupyter_lab",
```
jafreck: Sorry to be super pedantic, but I think this might be better as `jupyterlab` since they brand it as one word. I think we should also make plugins case insensitive (not in this pr, just in general).

Issue #424, opened 2018-02-28
Plugin setup does not respect `worker_on_mater`


jafreck:
https://github.com/Azure/aztk/blob/216f63dd64c630754150b64e43f9e05a2bb8d324/node_scripts/install/install.py#L27

PR #461, opened 2018-03-26
Feature

timotheeguerin:


jafreck:
This PR should also probably address #424

timotheeguerin:
Oh yeah think I already solve that one actually

jafreck:
This still need docs. I started doing some plugin docs in #467 which can be a starting point. Also, it seems like `ssh.yaml` is redundant now with plugins so we might want to remove it entirely.

aztk/internal/cluster_data/node_data.py
```diff
@@ -147,7 +146,8 @@ def _add_plugins(self):
                     execute='{0}/{1}'.format(plugin.name, plugin.execute),
                     args=plugin.args,
                     env=plugin.env,
-                    runOn=plugin.run_on.value,
+                    target=plugin.target.value,
+                    target_role=plugin.target_role.value,
```
paselem: This is a breaking change. All existing plugins use the runOn field which breaks them. This change should update all existing plusings and maybe deprecate runOn instead of remove it for now?

jafreck: Since plugins don't exist in the latest release (v0.6.0), this shouldn't be a breaking change. I think going forward we should limit deprecation to the cross release changes. 

Anyone using plugins today is using master, and should not expect the same level of stability as if they were upgrading from a previous minor release.

docs/51-define-plugin.md
```diff
@@ -0,0 +1,75 @@
...
+| `target`     | optional  | PluginTarget                    | Define where the execute script should be running. Potential values are `PluginTarget.SparkContainer(Default)` and `PluginTarget.Host`                          |
+| `taget_role` | optional  | PluginTargetRole                | If the plugin should be run only on the master worker or all. You can use environment variables(See below to have different master/worker config)               |             |
+
+### `PluginFile`
+| Name         | Required? | Type | Description                                                                  |
+|--------------|-----------|------|------------------------------------------------------------------------------|
+| `target`     | required  | str  | Where the file should be dropped relative to the plugin working directory    |
+| `local_path` | required  | str  | Path to the local file you want to upload(Could form the plugins parameters) |
+
+### `TextPluginFile`
```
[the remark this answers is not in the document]
timotheeguerin: Well text plugin file can then take a `string` not only a `string.io` then there is no way to know if the string is the content or the path to the file

aztk/node_scripts/setup_host.sh
```diff
@@ -2,19 +2,21 @@
 
 # Entry point for the start task. It will install all dependencies and start docker.
 # Usage:
-# setup_node.sh [container_name] [gpu_enabled] [docker_repo] [docker_cmd]
+# setup_host.sh [container_name] [gpu_enabled] [docker_repo] [docker_cmd]
```
jafreck: This needs to be updated to `#setup_host.sh [container_name] [repo_name]`. I keeping docker in the name of the variable for the docker repo is also easier to read. Maybe `docker_repo_name` is best?

aztk/node_scripts/install/spark_container.py
```diff
@@ -0,0 +1,72 @@
...
+    cmd.pass_env('SPARK_WORKER_UI_PORT')
+    cmd.pass_env('SPARK_CONTAINER_NAME')
+    cmd.pass_env('SPARK_SUBMIT_LOGS_FILE')
+    cmd.pass_env('SPARK_JOB_UI_PORT')
+
+    cmd.open_port(8080)       # Spark Master UI
+    cmd.open_port(7077)       # Spark Master
+    cmd.open_port(7337)       # Spark Shuffle Service
+    cmd.open_port(4040)       # Job UI
+    cmd.open_port(18080)     # Spark History Server UI
```
jafreck: Don't think this is the right place to do it, but we should only open the history server port if we detect that the user has configured to cluster to have it enabled (i.e. the proper keys are set in spark-defaults.conf when the cluster is created).

Otherwise, users might expect history server to be up when they have no configured their spark-defaults.conf correctly.

aztk/node_scripts/setup_host.sh
```diff
@@ -30,7 +32,13 @@ if ! host $HOSTNAME ; then
     echo $(hostname -I | awk '{print $1}') $HOSTNAME >> /etc/hosts
 fi
 
-if [ $gpu_enabled == "True" ]; then
+# Install docker-compose
```
jafreck: this isn't actually used in this PR. Can we comment it out/remove it until it is since this will add to cluster startup time?

aztk/node_scripts/install/spark_container.py
timotheeguerin: well this just opens the port on the docker container this shouldn't hurt. Though we could only display it in the ssh login if enabled. Maybe we could have the history server as a plugin too

aztk/node_scripts/setup_host.sh
timotheeguerin: Pablo is using it in his plugin. This allow plugins right now to use docker compose

aztk/utils/ssh.py
```diff
@@ -19,6 +17,7 @@ def connect(hostname,
             username=None,
             password=None,
             pkey=None):
+    import paramiko
```
jafreck: why is this necessary?

timotheeguerin: it was being imported on the cluster and we don;t install it on the cluster. Other options is to install on the cluster

jafreck: when/why was it being imported on the cluster?

timotheeguerin: dependency with module 

  File "/mnt/batch/tasks/startup/wd/aztk/node_scripts/main.py", line 2, in <module>
    import aztk.spark
  File "/mnt/batch/tasks/startup/wd/aztk/spark/__init__.py", line 2, in <module>
    from .client import Client
  File "/mnt/batch/tasks/startup/wd/aztk/spark/client.py", line 5, in <module>
    from aztk.client import Client as BaseClient
  File "/mnt/batch/tasks/startup/wd/aztk/client.py", line 15, in <module>
    import aztk.utils.ssh as ssh_lib
  File "/mnt/batch/tasks/startup/wd/aztk/utils/ssh.py", line 12, in <module>
    import paramiko
ImportError: No module named 'paramiko'

PR #464, opened 2018-03-27
Bug: fix aztk cluster submit paths, imports

### LINKED ISSUE #463: cluster submit broken since storage refactor
command_builder no longer copied to $DOCKER_WORKING_DIR, which breaks the cluster_submit command.

submit.py also moved to a new location.

jafreck:
Fix #463

### DIFF
--- a/aztk/node_scripts/docker_main.sh
+++ b/aztk/node_scripts/docker_main.sh
@@ -28,7 +28,8 @@ done
 echo "Starting setup using Docker"
 
 $(pyenv root)/versions/$AZTK_PYTHON_VERSION/bin/pip install -r $(dirname $0)/requirements.txt
-export PYTHONPATH=$PYTHONPATH:$DOCKER_WORKING_DIR
+
+echo 'export PYTHONPATH=$PYTHONPATH:$DOCKER_WORKING_DIR' >> ~/.bashrc
 
 echo "Running main.py script"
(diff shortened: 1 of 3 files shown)


aztk/node_scripts/docker_main.sh
```diff
@@ -28,7 +28,8 @@ done
 echo "Starting setup using Docker"
 
 $(pyenv root)/versions/$AZTK_PYTHON_VERSION/bin/pip install -r $(dirname $0)/requirements.txt
-export PYTHONPATH=$PYTHONPATH:$DOCKER_WORKING_DIR
+
+echo 'export PYTHONPATH=$PYTHONPATH:$DOCKER_WORKING_DIR' >> ~/.bashrc
```
timotheeguerin: don't you need to source bashrc now? or write the same line above

jafreck: yeah sourcing bash.rc is done in submit.py now: [here](https://github.com/Azure/aztk/pull/464/files#diff-05f414f140be1c8a088b89ddb5502a12R84)

timotheeguerin: yeah but don't we also need it here for docker_main

jafreck: I don't think so, at least not until we do #465 but it won't hurt to do now.

PR #470, opened 2018-04-02
Bug: line endings, add gitattributes file

jafreck:


docs/70-jobs.md
```diff
@@ -1,175 +1,175 @@
...
-
-
-### Stopping a Job's Application
-To stop an application that is running or going to run on a Job:
-
-```sh
-aztk spark job stop-app --id <your_job_id> --name <your_application_name>
-```
+# Jobs
+In the Azure Distributed Data Engineering Toolkit, a Job is a serverless entity that runs applications and records application output. A Job will manage the full lifecycle of the infrastructure so you do not have to. This document describes how to create and use AZTK Jobs.
```
paselem: "a Job is a serverless entity that..." That is not exactly true, right? Perhaps we should say "a Job is an entity that can run against a cluster that is automatically manged by the service. It can run applications and records outputs."

PR #475, opened 2018-04-03
Feature: getting started script

### LINKED ISSUE #354: Helper script to create Service Principal, Storage Account, Batch account
The script should:
1. Donwload and install the Azure CLI
2. use azure cli to login
3. create a service principal
4. create a storage account, and give service principal access
5. create a batch account and give service principal access
6. validate success
7. save secrets to `.aztk/secrets.yaml`
8. delete azure cli

The user should provide azure login credentials, but should not have to know or provide any other information.

The user should be able to configure service principal, batch account and storage account values. Default values for all should be configured.

This can be executed with an aztk cli command. Something like `aztk init`.

jafreck:
Fix #354


README.md
```diff
@@ -16,28 +16,32 @@ This toolkit is built on top of Azure Batch but does not require any Azure Batch
 - Ability to run _spark submit_ directly from your local machine's CLI
 
 ## Setup
+
 1. Clone the repo
```
paselem: Are we comfortable enough to just pip install? Technically it should be the same thing, right??

account_setup.py
```diff
@@ -0,0 +1,348 @@
...
+    except CloudError as e:
+        pass
+
+    if virtual_network:
+        confirmation_prompt = "A virtual network with the same name ({}) was found. \n"\
+                             "Please note that the existing address space and subnets may be changed or destroyed. \n"\
+                             "Do you want to use this virtual network? (y/n): ".format(virtual_network_name)
+        deny_error = AccountSetupError("Virtual network already exists, not recreating.")
+        unrecognized_input_error = AccountSetupError("Input not recognized.")
+        prompt_for_confirmation(confirmation_prompt, deny_error, unrecognized_input_error)
```
paselem: Never used this before, but does this exit the method?

account_setup.py
```diff
@@ -0,0 +1,348 @@
...
+        :param principal_id: str
+    """
+    authorization_client = AuthorizationManagementClient(credentials, subscription_id)
+    role_name = 'Contributor'
+    roles = list(authorization_client.role_definitions.list(
+        scope,
+        filter="roleName eq '{}'".format(role_name)
+    ))
+    contributor_role = roles[0]
+    for i in range(10):
```
paselem: What are you doing here? Is this trying to assign the roles on a retry loop?

account_setup.py
```diff
@@ -0,0 +1,348 @@
...
+
+
+if __name__ == "__main__":
+    print("\nGetting credentials.")
+    # get credentials and tenant_id
+    creds, subscription_id = credentials.get_azure_cli_credentials()
+    subscription_client = SubscriptionClient(creds)
+    tenant_ids = [tenant.id for tenant in subscription_client.tenants.list()]
+    if len(tenant_ids) != 1:
+        raise AccountSetupError("More than one tenant configured on your subscription.")
```
paselem: Is there nothing the user can do here? I'm not even sure what this error means myself...

account_setup.py
```diff
@@ -0,0 +1,348 @@
...
+
+    print("Input the desired names and values for your Azure resources. "\
+          "Default values are provided in the brackets. "\
+          "Hit enter to use default.")
+    kwargs = {
+        "region": prompt_with_default("Azure Region", DefaultSettings.region),
+        "resource_group": prompt_with_default("Resource Group Name", DefaultSettings.resource_group),
+        "storage_account": prompt_with_default("Storage Account Name", DefaultSettings.storage_account),
+        "batch_account": prompt_with_default("Batch Account Name", DefaultSettings.batch_account),
+        # "virtual_network_name": prompt_with_default("Virtual Network Name", DefaultSettings.virtual_network_name),
```
paselem: un-comment?

paselem: maybe this could be a 'leave blank for no vnet'

account_setup.py
```diff
@@ -0,0 +1,348 @@
...
+        :param **region: str
+    """
+    resource_client = ResourceManagementClient(credentials, subscription_id)
+    resource_client.resource_groups.list()
+    resource_group = resource_client.resource_groups.create_or_update(
+        resource_group_name=kwargs.get("resource_group", DefaultSettings.resource_group),
+        parameters={
+            'location': kwargs.get("region", DefaultSettings.region),
+        }
+    )
```
paselem: Will this fail w/ a good error msg is the user provides bad inputs? E.g. region == westuss

docs/01-getting-started-script.md
```diff
@@ -0,0 +1,44 @@
+# Getting Started Script
+
+The provided account setup script creates and configures all of the required Azure resources.
+
+The script will create and configure the following resources:
+- Resource group
+- Storage account
```
paselem: Should we add a vnet?
```sh
Virtual Network (optional)
```

README.md
jafreck: technically, yes. But I elected not to make that change here until our release on pip is "official". I can do it now if you'd prefer.

account_setup.py
jafreck: it will raise an error if either the user declines, or if in 3 tries the input is not recognized. Otherwise this returns none and just continues.

account_setup.py
jafreck: For some reason, the GraphRbacManagementClient's service_principal.create() method does not return an AzureOperationPoller to wait on (ala most of the other clients) nor does it block until the object is created server side. It just returns an object which may or may not have been created server side yet.

So I need to wait to give permissions here until the service principal is actually created.

account_setup.py
jafreck: This means that you're subscription has more than one tenant configured (don't know myself why that is possible/desired). I'm not sure what should be done here, or if we should bother. This seems like an extremely advanced case.

account_setup.py
jafreck: I left this commented since we don't (currently) have a way to output the subnet_id since it is in cluster.yaml and everything else is in secrets.yaml.

docs/01-getting-started-script.md
jafreck: Possibly, but I'm not sure how we'd do output. We could potentially just leave this out until we have "environments"

account_setup.py
jafreck: Right now this will just throw whatever error the ResourceManagementClient throws. I think we should probably catch this error, print a link to docs on azure regions, do a 3 attempt retry loop.

jafreck: This is the current error: 
```
  File "account_setup.py", line 336, in <module>
    resource_group_id = create_resource_group(creds, subscription_id, **kwargs)
  File "account_setup.py", line 79, in create_resource_group
    'location': kwargs.get("region", DefaultSettings.region),
  File "/home/jake/.local/lib/python3.5/site-packages/azure/mgmt/resource/resources/v2017_05_10/operations/resource_groups_operations.
py", line 147, in create_or_update
    raise exp
msrestazure.azure_exceptions.CloudError: Azure Error: LocationNotAvailableForResourceGroup
Message: The provided location 'westuss' is not available for resource group. List of available regions is 'centralus,eastasia,southea
stasia,eastus,eastus2,westus,westus2,northcentralus,southcentralus,westcentralus,northeurope,westeurope,japaneast,japanwest,brazilsout
h,australiasoutheast,australiaeast,westindia,southindia,centralindia,canadacentral,canadaeast,uksouth,ukwest,koreacentral,koreasouth,f
rancecentral,eastus2euap,centraluseuap'.
```

account_setup.py
```diff
@@ -0,0 +1,369 @@
...
+        "resource_group": prompt_with_default("Resource Group Name", DefaultSettings.resource_group),
+        "storage_account": prompt_with_default("Storage Account Name", DefaultSettings.storage_account),
+        "batch_account": prompt_with_default("Batch Account Name", DefaultSettings.batch_account),
+        # "virtual_network_name": prompt_with_default("Virtual Network Name", DefaultSettings.virtual_network_name),
+        # "subnet_name": prompt_with_default("Subnet Name", DefaultSettings.subnet_name),
+        "application_name": prompt_with_default("Active Directory Application Name", DefaultSettings.application_name),
+        "application_credential_name": prompt_with_default("Active Directory Application Credential Name", DefaultSettings.resource_group),
+        "service_principal": prompt_with_default("Service Principal Name", DefaultSettings.service_principal)
+    }
+    print("Creating the Azure resources.")
```
paselem: I just went through this again and parts of this can take a while, especially getting the service principal. I would recommend adding more print statements in this piece of code make it feel like progress is being made.

account_setup.py
```diff
@@ -0,0 +1,369 @@
...
+            "tenant_id": tenant_id,
+            "client_id": application_id,
+            "credential": application_credential,
+            # "subnet_id": subnet_id,
+            "batch_account_resource_id": batch_account_id,
+            "storage_account_resource_id": storage_account_id
+        }
+    )
+
+    print("\n# Copy the following into your .aztk/secrets.yaml file\n", secrets)
```
paselem: This is still printing with a space. I believe this will work
```python
 print("\n# Copy the following into your .aztk/secrets.yaml file\n{}".format(secrets))
```

PR #497, opened 2018-04-18
Feature: Readthedocs support

### LINKED ISSUE #496: Setup readthedocs
We are currently missing good documentation. Setting up readthedocs with docstring parsing with sphinx should help

timotheeguerin:
fix #496
http://aztk.readthedocs.io/en/feature-readthedocs/
image


jafreck:
In the SDK docs, I think the only package we should have is the `aztk.spark` package with only `aztk.spark.models` and `aztk.spark.client` underneath it (not utils). Those should be the only public facing modules.

docs/70-jobs.md
```diff
@@ -13,25 +13,25 @@ Creating a Job starts with defining the necessary properties in your `.aztk/job.
 Each Job has one or more applications given as a List in Job.yaml. Applications are defined using the following properties:
 ```yaml
   applications:
-    - name:
```
paselem: I thought we decided to leave in white spaces in yaml since it is required.

timotheeguerin: Hhm i guess its the auto formatting that removed those

requirements.txt
```diff
@@ -13,6 +13,14 @@ pylint==1.8.2
 pytest==3.1.3
 pytest-xdist==1.22.0
 twine==1.9.1
+docker==3.2.1
```
jafreck: why is this here?

PR #521, opened 2018-04-27
Bug: azure file share not being shared with container

### LINKED ISSUE #520: Azure files share not being shared with container
azure file shares aren't being passed to the docker command built on the host

jafreck:
Fix #520

### DIFF
--- a/aztk/node_scripts/install/install.py
+++ b/aztk/node_scripts/install/install.py
@@ -41,6 +41,7 @@ def setup_host(docker_repo: str):
 
     cluster_conf = read_cluster_config()
 
+    #TODO share azure file mounts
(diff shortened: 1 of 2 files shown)


aztk/node_scripts/install/install.py
```diff
@@ -41,6 +41,7 @@ def setup_host(docker_repo: str):
 
     cluster_conf = read_cluster_config()
 
+    #TODO share azure file mounts
```
timotheeguerin: ?

jafreck: I mean we should eventually share just the file shares that are specified, instead of all /mnt

Issue #527, opened 2018-04-30
Don't schedule driver to run on low priority instances


jaley:
I mentioned in #375 that it makes sense for us to run the driver on the master instance, because the master has capacity and won't be doing much else.

I noticed that currently, the driver may be scheduled to run on a low priority instance, which is causing stability problems on our ETL. Though I mentioned in the other ticket that I don't think it makes good sense to run the driver in client mode on an instance that's also running a worker, I think it's actually broken to schedule the driver to run on low priority instances when regular instances exist in the cluster.

The behaviour right now, if the instance is pre-empted, is that the whole job fails. For a big job, it's basically just as bad as losing the master, which we already have set to prefer non-preemptible instances.

timotheeguerin:
We can add functionality that it should run on a dedicated(or any node you'd want) as long as there is a available spot(From max task per node restriction on Batch which is x4 the number of cores of a node)
This should be a minor change, for the run on master to guarantee this is going to work all the time it is a bigger item which we still need to think about. Probably needs some options on where you want the driver to run (master, not master, dedicated, etc.)

jaley:
Yeah, this is the biggest stability issue we have right now. Scheduling the driver to run on a machine that's busy running workers causes big problems.

Several hours and 50TB or so into a large ETL job, the driver dies and the job restarts with a whole new retry from the beginning. I check the driver logs and see errors that a task has failed 4 times so it's giving up. Check the executor logs where the failing task was running, and I see something like this:

```
2018-04-30 16:06:55 WARN  Executor:87 - Issue communicating with driver in heartbeater
org.apache.spark.rpc.RpcTimeoutException: Futures timed out after [10 seconds]. This timeout is controlled by spark.executor.heartbeatInterval
	at org.apache.spark.rpc.RpcTimeout.org$apache$spark$rpc$RpcTimeout$$createRpcTimeoutException(RpcTimeout.scala:47)
	at org.apache.spark.rpc.RpcTimeout$$anonfun$addMessageIfTimeout$1.applyOrElse(RpcTimeout.scala:62)
	at org.apache.spark.rpc.RpcTimeout$$anonfun$addMessageIfTimeout$1.applyOrElse(RpcTimeout.scala:58)
	at scala.runtime.AbstractPartialFunction.apply(AbstractPartialFunction.scala:36)
	at org.apache.spark.rpc.RpcTimeout.awaitResult(RpcTimeout.scala:76)
[...]

PR #529, opened 2018-04-30
Feature: Increase max task per node

timotheeguerin:
Hard coded to 4 for now because we need to know the number of cores per vm size which would require getting the vm size specs

#527


aztk/client.py
```diff
@@ -108,6 +108,9 @@ def __create_pool_and_job(self, cluster_conf: models.ClusterConfiguration, softw
             start_task=start_task,
             enable_inter_node_communication=True if not cluster_conf.subnet_id else False,
             max_tasks_per_node=4,
+            task_scheduling_policy=batch_models.TaskSchedulingPolicy(
+                node_fill_type=batch_models.ComputeNodeFillType.pack,
```
paselem: Should we postpone until we know the service can handle this correctly?

jafreck: This is not a fix for anything, but it is better than what we have today. In the simple case, the underlying issue will not be present. Currently the second submitted application is not correctly handled. This delays the issue until the 5th concurrent application at least.

darylmsft:
Is “4” always going to work?

d

jafreck:
@darylmsft 4 will always work since the minimum number of cores is 1 and batch allows 4 tasks per core per vm.

Issue #498, opened 2018-04-19
Configure the CUDA BLAS libs for spark


jafreck:
Allows sparkml to run on GPUs:
- build spark with BLAS support
https://github.com/Mega-DatA-Lab/SpectralLDA-Spark/wiki/Compile-Spark-with-Native-BLAS-LAPACK-Support
- install BLAS
https://gist.github.com/dusenberrymw/d24fcefcd245cfacbc60d3b4caccc560

jafreck:
Added by #539

PR #539, opened 2018-05-03
Feature: nvBLAS and OpenBLAS plugin

jafreck:
#498


docker-image/gpu/spark1.6.3/Dockerfile
```diff
@@ -1,4 +1,159 @@
-FROM  aztk/spark:v0.1.0-spark1.6.3-base
```
paselem: why not build on top of this base image any more?

jafreck: There is a difference between base and gpu images in the Spark building process now (the added -Pnetlib-lgpl). We could build all images with that profile -- that might be the best option here.

aztk/spark/models/plugins/nvblas/configuration.py
```diff
@@ -0,0 +1,18 @@
+import os
+from aztk.models.plugins.plugin_configuration import PluginConfiguration, PluginPort, PluginTargetRole
+from aztk.models.plugins.plugin_file import PluginFile
+from aztk.utils import constants
+
+dir_path = os.path.dirname(os.path.realpath(__file__))
+
+class NvBLASPlugin(PluginConfiguration):
+    def __init__(self):
```
timotheeguerin: use a function here instead of a class as we talked

```
def NvBLASPlugin():
    return PluginConfiguration(...)
```

aztk/spark/models/plugins/openblas/configuration.py
```diff
@@ -0,0 +1,18 @@
+import os
+from aztk.models.plugins.plugin_configuration import PluginConfiguration, PluginPort, PluginTargetRole
+from aztk.models.plugins.plugin_file import PluginFile
+from aztk.utils import constants
+
+dir_path = os.path.dirname(os.path.realpath(__file__))
+
+class OpenBLASPlugin(PluginConfiguration):
```
timotheeguerin: Same

PR #540, opened 2018-05-04
Feature: Disable scheduling on group of nodes

### LINKED ISSUE #375: Make driver always execute on the Spark master

### LINKED ISSUE #527: Don't schedule driver to run on low priority instances

timotheeguerin:
fix #527 (Don't schedule on low pri nodes)
fix #375

Nodes that should not run the driver will have scheduling disabled

Todo: 
- [x] Handle when no nodes can match criteria
- [x] Docs for the new setting
- [x] Check job submission works
- [x] Tests


aztk/models/models.py
```diff
@@ -6,7 +6,7 @@
 from aztk.models.plugins import PluginConfiguration
 from aztk.internal import ConfigurationBase
 from .toolkit import Toolkit
-
+from .scheduling_target import SchedulingTarget
 
 class FileShare:
```
jafreck: insert second new line here

aztk/node_scripts/install/node_scheduling.py
```diff
@@ -0,0 +1,60 @@
...
+
+    node = batch_client.compute_node.get(pool_id, node_id)
+    if node.scheduling_state == batch.models.SchedulingState.disabled:
+        batch_client.compute_node.disable_scheduling(pool_id, node_id)
+        log.info("Enabled task scheduling for this node")
+    else:
+        log.info("Task scheduling is already enabled for this node")
+
+
+def setup_node_scheduling(
```
jafreck: Is this backwards compatible? Or is this breaking for existing clusters?

aztk/spark/models/models.py
```diff
@@ -202,30 +207,36 @@ def __init__(self, cloud_task: batch_models.CloudTask):
...
 
 
-class JobConfiguration:
+class JobConfiguration(ConfigurationBase):
     def __init__(
             self,
-            id,
-            applications,
-            vm_size,
+            id = None,
```
jafreck: The reason these values had no default is because they are required. Why are we defaulting to `None`? To facilitate merging?

timotheeguerin: Yeah this means we cant specify the default object, though I guess it could set it to none

aztk/node_scripts/install/node_scheduling.py
timotheeguerin: This shouldn't i guess, the actual scheduling shouldn't change

aztk/node_scripts/core/config.py
```diff
@@ -28,7 +28,7 @@
 cluster_id = os.environ.get("AZTK_CLUSTER_ID")
 pool_id = os.environ["AZ_BATCH_POOL_ID"]
 node_id = os.environ["AZ_BATCH_NODE_ID"]
-is_dedicated = os.environ["AZ_BATCH_NODE_IS_DEDICATED"]
+is_dedicated = os.environ.get("AZ_BATCH_NODE_IS_DEDICATED") == "true"
```
jafreck: why is this `os.environ.get()` and the above are not?

aztk/node_scripts/main.py
```diff
@@ -18,4 +20,5 @@ def run():
 
 
 if __name__ == "__main__":
+    logger.setup_logging()
```
jafreck: This isn't used anywhere it seems. Is this just for use later?

aztk/spark/client.py
```diff
@@ -21,10 +21,8 @@ class Client(BaseClient):
     Args:
         secrets_config(aztk.spark.models.models.SecretsConfiguration): Configuration with all the needed credentials
     """
-    def __init__(self, secrets_config):
-        super().__init__(secrets_config)
 
-    def create_cluster(self, cluster_conf: models.ClusterConfiguration, wait: bool = False):
+    def create_cluster(self, configuration: models.ClusterConfiguration, wait: bool = False):
```
jafreck: I think we should not do this since this is a breaking change. In the sdk rewrite, we will make sure the parameter names are better. But for now, we should preserve backwards compatibility and then deprecate these methods.

aztk/spark/client.py
```diff
@@ -200,8 +199,9 @@ def cluster_download(self, cluster_id: str, source_path: str, destination_path:
     '''
         job submission
     '''
-    def submit_job(self, job_configuration):
+    def submit_job(self, configuration: models.JobConfiguration):
```
jafreck: same as above

aztk/node_scripts/main.py
timotheeguerin: There was some logging in the node script which were basically not being printed to the stdout. This I think should fix it

timotheeguerin: Yep it works, I just updated the one still using logging which don't work

aztk/node_scripts/core/config.py
timotheeguerin: we could switch them all, though those are batch variables so they should always be set

jafreck: [`AZ_BATCH_NODE_IS_DEDICATED` is also a Batch variable though](https://docs.microsoft.com/en-us/azure/batch/batch-compute-node-environment-variables)

timotheeguerin: yeah so should we switch them all to `.get` or leave them with `[]`

jafreck: I feel like they should all be `[]` since that will throw an error if the environment variable is not present. `.get` will not until the variable is used -- and that is more confusing. Setting those environment variables is part of the service contract, so we should be fine failing if the service doesn't set them.

PR #543, opened 2018-05-07
Feature: Models v2

### LINKED ISSUE #162: Invalid keys in yaml files should result in an error instead of passive failure

timotheeguerin:
fix #162 (Unkown keys in config files will raise error)
Define a new Model base class and fields to be able to define models as below
This will handle most of the Validation, merging, parsing from dict automatically
```python

class UserState(Enum):
    Creating = "creating"
    Ready = "ready"
    Deleting = "deleting"

class UserInfo(Model):
    name = fields.String()
    age = fields.Integer()

class User(Model):
    info = fields.Model(UserInfo)
    enabled = fields.Boolean(default=True)
    state = fields.Enum(UserState, default=UserState.Ready)

# This is a valid way to create(from dict)
user = User(
        info=dict(
            name="Highlander",
            age=800,
        ),
        enabled=False,
        state="deleting",
    )
```

Todo: 

- [x] Handle list merging(Allow append, replace options)
- [x] Handle nested object merging(Allow merge, replace options)
- [x] Handle list parsing and validation correctly
- [x] Apply on existing models using configuration base
- [x] update writing model docs


.vscode/settings.json
```diff
@@ -16,5 +16,6 @@
     "--style=.style.yapf"
   ],
   "python.venvPath": "${workspaceFolder}/.venv/",
-  "python.pythonPath": "${workspaceFolder}/.venv/Scripts/python.exe"
+  "python.pythonPath": "${workspaceFolder}/.venv/Scripts/python.exe",
+  "python.unitTest.pyTestEnabled": true
```
darylmsft: is true the production value?

timotheeguerin: This setting is to have option to run the test one by one in vscode instead of using the cli, this is just for developement

aztk/models/cluster_configuration.py
```diff
@@ -0,0 +1,68 @@
...
+    """
+    Cluster configuration model
+
+    Args:
+        toolkit
+    """
+    cluster_id = fields.String()
+    toolkit = fields.Model(Toolkit)
+    size = fields.Integer(default=0)
+    size_low_pri = fields.Integer(default=0)
```
jafreck: what happens if `size` is set to 0 and `size_low_pri` is set to `None`?

aztk_cli/config.py
```diff
@@ -51,157 +46,42 @@ def _load_secrets_config(
...
-            password=docker_config.get('password'),
-        )
-
-    default_config = secrets_config.get('default')
-    # Check for ssh keys if they are provided
-    if default_config:
-        secrets.ssh_priv_key = default_config.get('ssh_priv_key')
-        secrets.ssh_pub_key = default_config.get('ssh_pub_key')
+    if 'default' in secrets_config:
+        deprecate("default key in secrets.yaml is deprecated. Place all child parameters directly at the root")
```
jafreck: I feel like this is a little confusing. Maybe we should point to docs on how to format secrets instead?

aztk/core/models/fields.py
```diff
@@ -0,0 +1,241 @@
...
+            if self.model and isinstance(item, collections.MutableMapping):
+                item = self.model(**item)
+            result.append(item)
+        return result
+
+    def merge(self, instance, value):
+        if value is None:
+            value = []
+
+        if self.merge_strategy == ListMergeStrategy.Append:
```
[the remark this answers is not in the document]
timotheeguerin: The other one is replace which is the default

PR #549, opened 2018-05-10
Feature: Support passing of remote executables via aztk spark cluster submit

### LINKED ISSUE #536: Support passing of remote executables via `aztk spark cluster submit`
As I understand it, the `aztk spark cluster submit` command uploads the given executable to the storage account associated with the cluster and then runs `spark submit` using the remote URI of the executable.

We have a use-case where the client does not have/need the executable jar and it is already accessible to the cluster at another storage location. We're currently faced with the situation of having to download the jar from one remote location in order to simply upload it to another.

What do you think about having an option to supply a remove path instead of a local executable? This is be done with an additional flag such as "--skip-executable-upload". For example:

```
aztk spark cluster submit \
    --id <name_of_spark_cluster> \
    --name <name_of_spark_job> \
    <remote_executable_path> \
    <executable_params> \
    --skip-executable-upload
```

lachiemurray:
fix #536

This change would allow the user to submit an app hosted at a remote location:

`aztk spark cluster submit [...] wasbs://path@remote/location.jar --skip-app-upload`

Or a location local to the docker container:

`aztk spark cluster submit [...] /local/to/docker_container.jar --skip-app-upload`

As well as supporting the current behaviour where the app is uploaded automatically

`aztk spark cluster submit [...] /local/absolute_path.jar`

The change involves moving the appending of "$AZ_BATCH_TASK_WORKING_DIR" to the application path to the client and only doing so if the "--skip-app-upload" flag is not set (note: the env variable is not resolved at this point since it is not known to the client). On the node, the now path is now un-altered.

Tested manually by:

* Submitting a job without the --skip-app-upload flag (current behaviour) :heavy_check_mark: 
* Submitting a job with the --skip-app-upload flag and using a wasb path to a remotely hosted jar :heavy_check_mark:
* Submitting a job with the --skip-app-upload flag and using a path to a jar that was copied to all nodes using `aztk spark cluster copy` :heavy_check_mark:

Seems like the integration tests are broken at the moment so haven't added any automated tests yet, happy to though if required.

### DIFF
--- a/aztk/node_scripts/submit.py
+++ b/aztk/node_scripts/submit.py
@@ -113,7 +113,9 @@ def __app_submit_cmd(
     if executor_cores:
         spark_submit_cmd.add_option('--executor-cores', str(executor_cores))
 
-    spark_submit_cmd.add_argument(app + ' ' + ' '.join(['\'' + str(app_arg) + '\'' for app_arg in (app_args or [])]))
+    spark_submit_cmd.add_argument(
+        os.path.expandvars(app) + ' ' + 
+        ' '.join(['\'' + str(app_arg) + '\'' for app_arg in (app_args or [])]))
 
     with open("spark-submit.txt", mode="w", encoding="UTF-8") as stream:
         stream.write(spark_submit_cmd.to_str())
--- a/aztk/node_scripts/submit.py
+++ b/aztk/node_scripts/submit.py
@@ -113,9 +113,7 @@ def __app_submit_cmd(
     if executor_cores:
         spark_submit_cmd.add_option('--executor-cores', str(executor_cores))
 
-    spark_submit_cmd.add_argument(
-        os.environ['AZ_BATCH_TASK_WORKING_DIR'] + '/' + app + ' ' +
-        ' '.join(['\'' + str(app_arg) + '\'' for app_arg in (app_args or [])]))
+    spark_submit_cmd.add_argument(app + ' ' + ' '.join(['\'' + str(app_arg) + '\'' for app_arg in (app_args or [])]))
 
     with open("spark-submit.txt", mode="w", encoding="UTF-8") as stream:
         stream.write(spark_submit_cmd.to_str())
@@ -156,7 +154,7 @@ def recieve_submit_request(application_file_path):
 
     cmd = __app_submit_cmd(
         name=application['name'],
-        app=os.path.basename(application['application']),
+        app=application['application'],
         app_args=application['application_args'],
         main_class=application['main_class'],
         jars=application['jars'],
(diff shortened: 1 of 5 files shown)


aztk/node_scripts/submit.py
```diff
@@ -113,9 +113,7 @@ def __app_submit_cmd(
     if executor_cores:
         spark_submit_cmd.add_option('--executor-cores', str(executor_cores))
 
-    spark_submit_cmd.add_argument(
-        os.environ['AZ_BATCH_TASK_WORKING_DIR'] + '/' + app + ' ' +
-        ' '.join(['\'' + str(app_arg) + '\'' for app_arg in (app_args or [])]))
+    spark_submit_cmd.add_argument(app + ' ' + ' '.join(['\'' + str(app_arg) + '\'' for app_arg in (app_args or [])]))
```
lachiemurray: Currently just letting the $AZ_BATCH_TASK_WORKING_DIR be expanded when the spark submit command is executed but could be more explicit and do it here with os.path.expandvars

timotheeguerin:
So i am wondering if this is the right way of doing it. Maybe we should make it the other way in that you can give the path to a remote location or an existing path on the node(exe needs to be there) but if you want to upload your local file you provide the flag

lachiemurray:
Yes, I think I agree, but it's obviously more of a breaking change using "--local". I wonder whether it's a lot to do with the naming that makes it seem hacky though. Maybe "--remote" would seem more natural? 

Happy to give "--local" a go or rename to "--remote", just let me know. 

Do you have any advise regarding testing? Should I try to write an integration test despite them not running on travis at the moment?

timotheeguerin:
Yeah I think `--remote` makes more sense.
That might actually be better than using `--local` as I don't like the idea that you have to provide more settings when you are trying to get started and upload one of the samples or any local files

Will talk to @jafreck

aztk/node_scripts/submit.py
timotheeguerin: Sure that could be a good idea if we also log that command for debugging purposes

jafreck:
I agree with @lachiemurray here in regards to making local files default for cluster submission. I think that `--remote` is a better naming than `--skip-app-upload` as well.

Thanks for doing this @lachiemurray!

timotheeguerin:
okok lets rename to `--remote`, could you also provide docs for this flag in `20-spark-submit.md`

lachiemurray:
Thanks both! Will make the changes tomorrow

lachiemurray:
Renamed to remote and updated the docs, can you take another look @timotheeguerin @jafreck?

PR #569, opened 2018-05-21
Fix: getting started script reuse aad application

### LINKED ISSUE #562: Getting started script breaks if using existing AAD application
If using an existing application, the script does not add a new password credential, so authentication will not work with the returned secrets.

Also the getting started script has a few typos that should be corrected.

jafreck:
Fix #562

### DIFF
--- a/account_setup.py
+++ b/account_setup.py
@@ -189,6 +189,7 @@ def create_aad_user(credentials, tenant_id, **kwargs):
                 display_name=display_name,
                 password_credentials=[
                     PasswordCredential(
+                        start_date=datetime(2000, 1, 1, 0, 0, 0, 0, tzinfo=timezone.utc),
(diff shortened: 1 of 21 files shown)


account_setup.py
```diff
@@ -189,6 +189,7 @@ def create_aad_user(credentials, tenant_id, **kwargs):
                 display_name=display_name,
                 password_credentials=[
                     PasswordCredential(
+                        start_date=datetime(2000, 1, 1, 0, 0, 0, 0, tzinfo=timezone.utc),
```
timotheeguerin: shoudn't that be current time?

jafreck: I don't think it should matter, It will have the same effect.

We shouldn't have to include this line at all, this is only necessary because there is a bug in the graphrbac client if start_date isn't specified.

PR #570, opened 2018-05-22
Fixed path issues with jupyter and jupyter_lab plugins

edouardpoitras:
Signed-off-by: Edouard Poitras <eddie@quantumcyberdefence.com>

### DIFF
--- a/aztk/spark/models/plugins/jupyter_lab/jupyter_lab.sh
+++ b/aztk/spark/models/plugins/jupyter_lab/jupyter_lab.sh
@@ -9,12 +9,12 @@
 if  [ "$AZTK_IS_MASTER" = "true" ]; then
     conda install -c conda-force jupyterlab
 
-    PYSPARK_DRIVER_PYTHON="/opt/conda/bin/jupyter"
-    JUPYTER_KERNELS="/opt/conda/share/jupyter/kernels"
+    PYSPARK_DRIVER_PYTHON="/.pyenv/versions/${USER_PYTHON_VERSION}/bin/jupyter"
+    JUPYTER_KERNELS="/.pyenv/versions/${USER_PYTHON_VERSION}/share/jupyter/kernels"
 
     # disable password/token on jupyter notebook
     jupyter lab --generate-config --allow-root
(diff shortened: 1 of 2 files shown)


jafreck:
`pyenv` is no longer used in the latest docker images, to use the jupyter/jupyterlab plugins you will need to use the newer docker images with the minconda or anaconda environment.

To use the newer Dockerfiles, you will need to upgrade to aztk version 0.7.0+ and specify a toolkit block in your cluster.yaml (if using the cli), or modify your ClusterConfiguration object (if using the SDK). Y ou can follow the migration guide for more details, [here](http://aztk.readthedocs.io/en/latest/80-migration.html#to-0-7-0).

If you are using your own Dockerfiles, you will need to adhere to the new requirements, [here](http://aztk.readthedocs.io/en/latest/12-docker-image.html#using-a-custom-docker-image). For reference, this is the Dockerfile for the new [Spark 2.3.0 miniconda image](https://github.com/Azure/aztk/blob/master/docker-image/miniconda/spark2.3.0/base/Dockerfile), and here is the [Dockerfile for the Spark 2.3.0 base image](https://github.com/Azure/aztk/blob/master/docker-image/base/spark2.3.0/Dockerfile).

edouardpoitras:
OK, thanks Jacob.

aztk/spark/models/plugins/jupyter_lab/jupyter_lab.sh
```diff
@@ -42,7 +42,7 @@ if  [ "$AZTK_IS_MASTER" = "true" ]; then
     "env": {
         "SPARK_HOME": "$SPARK_HOME",
         "PYSPARK_PYTHON": "python",
-        "PYSPARK_SUBMIT_ARGS": "--master spark://$MASTER_IP:7077 pyspark-shell"
+        "PYSPARK_SUBMIT_ARGS": "--master spark://$AZTK_MASTER_IP:7077 pyspark-shell"
```
jafreck: This seems to actually be a bug, though. `$AZTK_MASTER_IP` is the correct environment variable.

PR #572, opened 2018-05-22
Feature: add node run command

### LINKED ISSUE #571: Add ability to run command on specific node

jafreck:
Fix #571


docs/10-clusters.md
```diff
@@ -86,6 +86,14 @@ aztk spark cluster run --id <your_cluster_id> "<command>"
 
 The command is executed through an SSH tunnel.
 
+### Run a command on a specific node in the cluster
+To run a command on all nodes in the cluster, run:
+```sh
+aztk spark cluster node-run --id <your_cluster_id> --node-id <your_node_id> "<command>"
```
timotheeguerin: exec or node-exec? Stays more in line with docker exec

aztk_cli/spark/endpoints/cluster/cluster_node_run.py
```diff
@@ -0,0 +1,39 @@
...
+
+def execute(args: typing.NamedTuple):
+    spark_client = aztk.spark.Client(config.load_aztk_secrets())
+    with utils.Spinner():
+        node_id, output = spark_client.node_run(args.cluster_id, args.node_id, args.command, args.internal)
+    print_execute_result(node_id, output)
+
+
+def print_execute_result(node_id, result):
+    print("-" * (len(node_id) + 6))
```
timotheeguerin: should use the log here i think

aztk/client.py
```diff
@@ -225,6 +225,32 @@ def __delete_user_on_pool(self, username, pool_id, nodes):
             futures = [exector.submit(self.__delete_user, pool_id, node.id, username) for node in nodes]
             concurrent.futures.wait(futures)
 
+    def __node_run(self, cluster_id, node_id, command, internal, container_name=None, timeout=None):
+        pool, nodes = self.__get_pool_details(cluster_id)
+        node = [node for node in nodes if node.id == node_id]
```
timotheeguerin: would use `next(node for node in nodes if node.id == node_id)` here instead of this and the if below https://stackoverflow.com/questions/9542738/python-find-in-list

aztk/client.py
```diff
@@ -225,6 +225,32 @@ def __delete_user_on_pool(self, username, pool_id, nodes):
...
+        if node:
+            node = node[0]
+        if internal:
+            node_rls = models.RemoteLogin(ip_address=node.ip_address, port="22")
+        else:
+            node_rls = self.__get_remote_login_settings(pool.id, node.id)
+        ssh_key = RSA.generate(2048)
+        ssh_pub_key = ssh_key.publickey().exportKey('OpenSSH').decode('utf-8')
+        try:
+            self.__create_user_on_node('aztk', pool.id, node.id, ssh_pub_key)
```
timotheeguerin: shouldn't this user have some kind of random id so it doesn't conflict

docs/10-clusters.md
jafreck: There is already a `cluster run` so I figured `cluster node-run` made the most sense here.

aztk/client.py
jafreck: maybe, but if the user already exists, then it is just deleted and recreated. So the only potential issue is if someone adds `aztk` as a user, and then runs a command. In that case, their created user would be deleted, so adding a randomly generated user id would probably be best here.

timotheeguerin: or running node-run twice at the same time. Not sure what deleting the user while being login does

jafreck: I believe nothing would happen while the session is open. However, still, making the user id generated would solve that.

jafreck:
On second thought, for this PR, maybe the CLI should just be `aztk spark cluster run --id <id> --node-id <tvm-xxx>` instead of an entirely seperate endpoint.

timotheeguerin:
Thought this, the problem is the same we talked about earlier where you might make a mistake and run on all nodes instead of a single one. Is it however a big deal?

aztk/client.py
```diff
@@ -4,16 +4,19 @@
 
 import azure.batch.models as batch_models
 import azure.batch.models.batch_error as batch_error
+from azure.batch.models import batch_error as batch_error
```
timotheeguerin: is this import the same as above?

aztk_cli/utils.py
```diff
@@ -449,3 +449,14 @@ def print_cluster_conf(cluster_conf: ClusterConfiguration, wait: bool):
...
+
+def log_execute_result(node_id, result):
+    log.info("-" * (len(node_id) + 4))
+    log.info("| %s |", node_id)
+    log.info("-" * (len(node_id) + 4))
+    if isinstance(result, Exception):
+        log.info("%s\n", result)
+    else:
+        for line in result:
+            print(line)
```
timotheeguerin: still a print here?

PR #576, opened 2018-05-23
Fix: add plugins to Spark ClusterConfiguration init

### LINKED ISSUE #575: aztk.spark.models.ClusterConfiguration does not take plugins in __init__
To specify plugins in the SDK, the Spark ClusterConfiguration should accept a list of plugins.

jafreck:
Fix #575


jafreck:
Fixed by #543

PR #577, opened 2018-05-24
Feature: pure python ssh

### LINKED ISSUE #296: Make ssh connection pure python
Remove ssh client dependency

jafreck:
Fix #296 

To do:
- [x] Add support for `--internal` flag


aztk/utils/ssh.py
```diff
@@ -133,3 +208,25 @@ def node_copy(node_id, source_path, destination_path, username, hostname, port,
...
+        threads = forward_ports(client=client, port_forward_list=port_forward_list)
+    except AztkError as e:
+        raise e
+
+    try:
+        import time
+        while True:
+            time.sleep(1)
+    except KeyboardInterrupt:
+        pass
```
timotheeguerin: doesn't this prevent ctrl + d

aztk/utils/ssh.py
```diff
@@ -8,10 +8,70 @@
...
 import socketserver as SocketServer
 import sys
+import threading
 from concurrent.futures import ThreadPoolExecutor
 
 from aztk.error import AztkError
-from . import helpers
+
+
+g_verbose = False
```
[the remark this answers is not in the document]
jafreck: yeah that would be better

aztk/utils/ssh.py
timotheeguerin: Can you add a comment about it so we don't forget why its needed

aztk/utils/ssh.py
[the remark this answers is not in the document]
jafreck: actually, since this is in the SDK, we don't current have a logger. Maybe that would be nice to have though.

aztk/utils/ssh.py
```diff
@@ -8,10 +8,70 @@
...
+    if g_verbose:
+        print(s)
+
+
+class ForwardServer(SocketServer.ThreadingTCPServer):
+    daemon_threads = True
+    allow_reuse_address = True
+
+# pylint: disable=no-member
+class Handler(SocketServer.BaseRequestHandler):
```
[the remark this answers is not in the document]
jafreck: I feel like this belongs here since this is the ssh tunnel request handler and this file is meant to have the utilities necessary to make ssh connections and commands.

aztk_cli/spark/endpoints/cluster/cluster_ssh.py
```diff
@@ -88,16 +73,72 @@ def print_plugin_ports(cluster_config: ClusterConfiguration):
...
+    plugin_ports = []
+    if configuration and configuration.plugins:
+        ports = [
+            PortForwardingSpecification(
+                port.internal,
+                port.public_port) for plugin in configuration.plugins for port in plugin.ports if port.expose_publicly
+        ]
+        plugin_ports.extend(ports)
+
+    print("Press ctrl+c to exit...")
```
timotheeguerin: shouldn't that be ctrl + d? or not as we are not giving shell access

aztk_cli/spark/endpoints/cluster/cluster_ssh.py
```diff
@@ -88,16 +73,72 @@ def print_plugin_ports(cluster_config: ClusterConfiguration):
...
+            connect=ssh_conf.connect,
+            internal=ssh_conf.internal)
+
+        if not ssh_conf.connect:
+            log.info("")
+            log.info("Use the following command to connect to your spark head node:")
+            log.info("\t%s", ssh_cmd)
+
+    except batch_error.BatchErrorException as e:
+        if e.error.code == "PoolNotFound":
```
timotheeguerin: I think we should have an enum(or const) for those error codes

aztk_cli/spark/endpoints/cluster/cluster_ssh.py
jafreck: We aren't giving shell access, the process just hangs so the ports stay open until killed (with `crl + c`).

aztk_cli/spark/endpoints/cluster/cluster_ssh.py
jafreck: agreed, but I think that should probably come in a single PR for the entire repo.

PR #581, opened 2018-05-25
Feature: add cluster list quiet flag, ability to compose with delete

### LINKED ISSUE #481: Add ability to delete multiple jobs/clusters at once
Would also be really nice to be able to do something like:

`aztk spark cluster list --id-only | aztk spark cluster delete --id` to delete all clusters.

### LINKED ISSUE #578: Add quiet flag to cluster list
Quiet should only output `cluster_id`'s.

This will help with composing CLI commands

jafreck:
Fix #578 
Fix #481


aztk_cli/utils.py
```diff
@@ -119,6 +119,11 @@ def print_clusters(clusters: List[models.Cluster]):
             )
         )
 
+
+def print_clusters_quiet(clusters: List[models.Cluster]):
+    print('\n'.join([str(cluster.id) for cluster in clusters]))
```
timotheeguerin: should we wait for the other PR and change to log.print

PR #582, opened 2018-05-25
Fix: optimize start task

### LINKED ISSUE #499: Optimize setup_node.sh
Currently there are a lot of unoptimized code in setup_node.py which can impact cluster startup time.

jafreck:
Fix #499


aztk/node_scripts/setup_host.sh
```diff
@@ -12,22 +12,32 @@ container_name=$1
...
-
-# Install docker
-echo "Installing Docker"
-curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
-add-apt-repository "deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable"
-apt-get -y update
-apt-get -y install docker-ce
-echo "Done installing Docker"
+time(
+    curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
```
timotheeguerin: we could maybe split those into different files so it makes it more readable

jafreck: split what into different files? the inside of the time statement?

timotheeguerin: Yeah

PR #593, opened 2018-06-05
Fix: --size-low-pri being ignored

### LINKED ISSUE #592: "cluster create" ignores the --size-low-pri command line parameter
The parser does not complain if you pass in --size-low-pri <SIZE_LOW_PRI>, but the behaviour is as if the parameter had not been passed in at all: The number of low-priority VMs is what was specified in cluster.yaml, or zero if not specified there.

timotheeguerin:
fix #592


aztk_cli/spark/endpoints/cluster/cluster_create.py
```diff
@@ -40,6 +43,10 @@ def execute(args: typing.NamedTuple):
     # read cluster.yaml configuartion file, overwrite values with args
```
mmduyzend: (While you're here, could you fix the spelling of "configuration" in the comment?)

Issue #547, opened 2018-05-08
Add apt-get install plugin


jafreck:

timotheeguerin:
I guess this one would need to run before the docker run optimally

jafreck:
I was thinking this would run in the container, I'm not sure I see why someone would want to run this on the host, but perhaps that would be a good feature as well.

PR #594, opened 2018-06-06
Conda, Apt-Get and Pip Install Plugins

brnleehng:
#545 (conda install plugin)
#546 (pip install plugin)
#547 (apt install plugin)


aztk/models/plugins/internal/plugin_manager.py
```diff
@@ -26,6 +26,9 @@ class PluginManager:
         tensorflow_on_spark=plugins.TensorflowOnSparkPlugin,
         openblas=plugins.OpenBLASPlugin,
         nvblas=plugins.NvBLASPlugin,
+        pip=plugins.PipPlugin,
+        apt_get=plugins.AptGetPlugin,
```
timotheeguerin: should we name this one "apt-get" or just "apt"?

aztk/spark/models/plugins/install/install.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+
+apt-get update
```
timotheeguerin: shouldn't that only be for the apt?

aztk/spark/models/plugins/install/install.sh
```diff
@@ -0,0 +1,8 @@
...
+
+for PACKAGE in "$@"
+do
+    eval $COMMAND $PACKAGE
```
timotheeguerin: also for apt I think as @jafreck noticed it might be more performant to run apt install -y package1 package2, etc instead of multiple apt instance
(I think you should be able to do the same for pip and conda)

aztk/spark/models/plugins/install/install.sh
jafreck: yeah, this is going to be refactored out. I think that having a single `install.sh` is fine so long as we do a check that $COMMAND is apt-get.

aztk/spark/models/plugins/install/conda/configuration.py
```diff
@@ -0,0 +1,18 @@
...
+def CondaPlugin(packages=None):
+    return PluginConfiguration(
+        name="conda",
+        target_role=PluginTargetRole.All,
+        execute="install.sh",
+        files=[
+            PluginFile("install.sh", os.path.join(dir_path, "..", "install.sh"))
+        ],
+        args=packages,
+        env=dict(COMMAND='conda install')
```
jafreck: Looks like the `-y` flag is necessary here to get around prompting. https://conda.io/docs/commands/conda-install.html

aztk/models/plugins/internal/plugin_manager.py
jafreck: technically, since we are calling `apt-get` and not `apt` I think this is fine as is.

jafreck: I think we should call these `apt_get_install`, `conda_install` and `pip_install` since that is more indicative of what the plugin does.

aztk/spark/models/plugins/install/install.sh
brnleehng: Will only do one instance

aztk/models/plugins/internal/plugin_manager.py
timotheeguerin: Though to compare, this is travis apt plugin https://docs.travis-ci.com/user/installing-dependencies/#Adding-APT-Packages

aztk/spark/models/plugins/install/install.sh
brnleehng: Yeah I'll move it to the full command

aztk/models/plugins/internal/plugin_manager.py
timotheeguerin: This also has an option to update or not before

aztk/spark/models/plugins/install/conda/configuration.py
brnleehng: WIll add the -y flag

aztk/models/plugins/internal/plugin_manager.py
jafreck: I can't think of a case where you wouldn't want to update.

brnleehng: Going to stick with the current convention of apt_get since the rest of the plugins are with '_'

PR #600, opened 2018-06-11
Fix: release v0.8.0

jafreck:


CHANGELOG.md
```diff
@@ -28,6 +21,20 @@
...
+* optimize start task (#582) ([e5e529a](https://github.com/jafreck/aztk/commit/e5e529a)), closes [#582](https://github.com/jafreck/aztk/issues/582)
+* remove deprecated vm_count call (#586) ([dbde8bc](https://github.com/jafreck/aztk/commit/dbde8bc)), closes [#586](https://github.com/jafreck/aztk/issues/586)
+* Remove old spark-defaults.conf jars (#567) ([8b8cd62](https://github.com/jafreck/aztk/commit/8b8cd62)), closes [#567](https://github.com/jafreck/aztk/issues/567)
+* set logger to stdout (#588) ([3f0c8f9](https://github.com/jafreck/aztk/commit/3f0c8f9)), closes [#588](https://github.com/jafreck/aztk/issues/588)
+* switch create user to pool wide (#574) ([49a890a](https://github.com/jafreck/aztk/commit/49a890a)), closes [#574](https://github.com/jafreck/aztk/issues/574)
+* switch from pycryptodome to pycryptodomex (#564) ([19dde42](https://github.com/jafreck/aztk/commit/19dde42)), closes [#564](https://github.com/jafreck/aztk/issues/564)
+* allow cluster config to be printed when no username has been set (#597) ([1cc71c7](https://github.com/jafreck/aztk/commit/1cc71c7)), closes [#597](https://github.com/jafreck/aztk/issues/597)
```
mmduyzend: Shouldn't these links be pointing to Azure/aztk rather than to jafreck/aztk?

aztk/models/cluster_configuration.py
```diff
@@ -101,7 +101,7 @@ def __validate__(self) -> bool:
             )
 
         if self.custom_scripts:
-            deprecate("Custom scripts are DEPRECATED.", "Use plugins instead. See https://aztk.readthedocs.io/en/v0.7.0/15-plugins.html.")
+            deprecate("0.9.0", "Custom scripts are DEPRECATED.", "Use plugins instead. See https://aztk.readthedocs.io/en/v0.7.0/15-plugins.html.")
```
mmduyzend: There are several links in the codebase and documentation to v0.7.0 of readthedocs, and also some links to the v0.7.0 tag on GitHub. Should these be updated?

CHANGELOG.md
```diff
@@ -2,18 +2,11 @@
 
 ##  0.8.0 (2018-06-07)
```
mmduyzend: It actually went out a few days later.

jafreck: you are right... we use an automated tool to generate the changelog, so not sure why this would be out of date. Thanks for pointing this out.

aztk/models/cluster_configuration.py
jafreck: yes, we can release a patch update to pitch these issues. Thanks for pointing this out.

CHANGELOG.md
jafreck: yes, you are correct. It seems like there is an issue with our changelog generator build. We will make sure that these issues are ironed out with a 0.8.1 patch update. Thanks for pointing this out.

jafreck:
@mmduyzend Our deployment relies on a changelog generator, and it seems like most of these issue are arising from there. We will make sure to fix the issue with the build and deploy a 0.8.1 patch update to fix.

aztk/models/cluster_configuration.py
mmduyzend: Maybe it would make sense for them to point to `latest` in the source code, and then to have the release automation fix them to a particular version?

PR #643, opened 2018-08-10
Feature: Example SDK additional functionality for parsing yaml config

atg-abhishek:
Parsing secrets.yaml config file to prevent duplicate entry of config vars when following the getting started tutorial


atg-abhishek:
Hey @jafreck would you be the right person to review and merge this PR? Thanks!

jafreck:
Hey @atg-abhishek, I am a bit conflicted about this change. That example file is supposed to show SDK usage, and is not meant to rely on any single location for the secrets.

Certainly, supplying a yaml file and parsing it is an option, but it is not the only one. This example is meant to be generic in that sense. I do see how this would be helpful for users who are currently using the CLI and want to try out the SDK, though.

@timotheeguerin do you have an opinion on this?

timotheeguerin:
Sorry missed that one. Yes do agree with @jafreck, this sample should be about how to get a aztk client and not how to handle loading the credentials which in the end is probably going to be very user specific. Loading the secrets from a yaml file might feel like restrict the user choice.

jafreck:
Thanks @timotheeguerin. I'm going to close this PR, then. @atg-abhishek Thanks for the suggestion, we appreciate contributions like this. Apologies, we aren't going to take this one -- we want to keep this sample generic.

atg-abhishek:
No worries about that!

PR #645, opened 2018-08-13
Feature: 0.9.0 deprecated code removal

jafreck:


aztk/models/cluster_configuration.py
```diff
@@ -41,37 +41,8 @@ class ClusterConfiguration(Model):
...
-
-        if 'vm_low_pri_count' in kwargs:
-            deprecate("vm_low_pri_count is deprecated for ClusterConfiguration.",
-                      "Please use size_low_priority instead.")
-            kwargs['size_low_priority'] = kwargs.pop('vm_low_pri_count')
-
         super().__init__(*args, **kwargs)
 
-    @property
-    @deprecated("0.9.0")
```
timotheeguerin: did we remove also the one we deprecated before but didn't add a version to it?

jafreck: I don't see any instances of non-versioned deprecations. I believe we added versions to all of these.

PR #653, opened 2018-08-24
Fix: spark roll back scheduling disable

jafreck:
Disabling scheduling_target feature because of Batch node rejoin pool bug.


docs/13-configuration.md
```diff
@@ -47,12 +47,14 @@ plugins:
 # Allow master node to also be a worker <true/false> (Default: true)
 # worker_on_master: true
 
-# Where do you want to run the driver <dedicated/master/any> (Default: dedicated if at least one dedicated node or any otherwise)
-# scheduling_target: dedicated
 
 # wait: <true/false>
 wait: true
 ```
+<!--- this goes about wait: true
```
timotheeguerin: ?

jafreck: This should read "this goes after `wait: true`". The reason this is here is because you can't comment out things inside a code block in markdown, so I moved the scheduling stuff outside the code block and into a comment. This is just meant to point to where it should go when it is uncommented.

PR #661, opened 2018-09-20
Feature: Spark scheduling target

jafreck:
Reintroduce `scheduling_target` with a direct-to-node implementation that does not rely on Batch scheduling.

Reqs:
- [x] Submit application to master node
- [x] Resource file ingress
- [x] Get application status
- [x] Get application log
- [x] output / error file egress
- [x] get application `exit_code`
- [x] get application execution time
- [x] Add integration test for scheduling_target

This implementation should be aligned as much as possible with the standard Batch Task scheduling feature to maximize code reuse.

Fix #670
Fix #527

PR #672, opened 2018-10-24
Feature: Spark retry docker pull

### LINKED ISSUE #659: Start task failure: error pulling image: 503 The server is busy
When provisioning a new cluster with more than 50 nodes we start to see a large proportion of start task failures (25+%) when pulling the docker image from an ACR.

```
error pulling image
configuration: received unexpected HTTP status: 503 The server is busy
```

I raised a support issue with the ACR team and they said we were being throttled and recommend that we attempt to retry pulling the image when we get a 503.

Note our image is quite large (~7 GB) which may be why we experience this issue whilst other do not.

Here are some of the ways I think we could mitigate this:

1) Retry pulling the image when 503 returned (as recommended by the ACR team)

2) Retry the entire start-task when any failure occurs (it looks like there is support for this in the [batch SDK](https://github.com/Azure/azure-sdk-for-python/blob/41e37c8a10876db40697a63e828ed7cafc19c7d6/azure-mgmt-batch/azure/mgmt/batch/models/start_task.py#L70)?)

3) Configure the docker daemon to pull fewer layers in parallel using the [max-concurrent-downloads](https://docs.docker.com/engine/reference/commandline/pull/#concurrent-downloads) option. I looked at whether this could be done using a plugin but I think plugins would be run too late?

What do you think the best approach would be? Can you recommend any others?

jafreck:
This retries docker pulls a maximum of 5 times with an exponential back-off. 

In addition to the start task retries introduced with #661, this should alleviate some issues with intermittent 
network errors when pulling docker images.

Fix #659

### DIFF
--- a/aztk/node_scripts/setup_host.sh
+++ b/aztk/node_scripts/setup_host.sh
@@ -42,8 +42,8 @@ install_prerequisites () {
 
 install_docker_compose () {
     echo "Installing Docker-Compose"
-    for i in {1..5}; do 
-        sudo curl -L https://github.com/docker/compose/releases/download/1.19.0/docker-compose-`uname -s`-`uname -m` -o /usr/local/bin/docker-compose && break || sleep 2; 
+    for i in {1..5}; do
+        sudo curl -L https://github.com/docker/compose/releases/download/1.19.0/docker-compose-`uname -s`-`uname -m` -o /usr/local/bin/docker-compose && break || sleep 2;
     done
     sudo chmod +x /usr/local/bin/docker-compose
     echo "Finished installing Docker-Compose"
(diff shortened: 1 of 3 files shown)


aztk/node_scripts/setup_host.sh
```diff
@@ -42,8 +42,8 @@ install_prerequisites () {
 
 install_docker_compose () {
     echo "Installing Docker-Compose"
-    for i in {1..5}; do 
-        sudo curl -L https://github.com/docker/compose/releases/download/1.19.0/docker-compose-`uname -s`-`uname -m` -o /usr/local/bin/docker-compose && break || sleep 2; 
+    for i in {1..5}; do
+        sudo curl -L https://github.com/docker/compose/releases/download/1.19.0/docker-compose-`uname -s`-`uname -m` -o /usr/local/bin/docker-compose && break || sleep 2;
```
timotheeguerin: add a echo here if it fails?

PR #677, opened 2018-10-29
Feature: ClusterState enum

jafreck:
Align ClusterState with the JobState and ApplicationState Enums.

### DIFF
--- a/aztk/models/cluster_state.py
+++ b/aztk/models/cluster_state.py
@@ -1,0 +1,8 @@
+from enum import Enum
+
+
+class ClusterState(Enum):
+    deleting = "deleting"
+    resizing = "resizing"
+    steady = "steady"
+    stopping = "stopping"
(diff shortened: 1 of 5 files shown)


aztk/models/cluster_state.py
```diff
@@ -0,0 +1,8 @@
+from enum import Enum
+
+
+class ClusterState(Enum):
+    deleting = "deleting"
+    resizing = "resizing"
+    steady = "steady"
+    stopping = "stopping"
```
timotheeguerin: `stopping` feels weird actually in this enum, could it be renamed to stoppingResize?

jafreck: Sure. Camel casing the Enum looks weird to me, though. Looking back at python docs, it seems like `Enum.SOME_NAME` is the idiomatic casing. I guess I should align to that for all.

jafreck: Actually, I'm going to stick with `Enum.some_value` since that is what the Batch SDK does.

PR #678, opened 2018-11-01
Include cluster creation time in output of cluster list and cluster get

emlyn:
When listing clusters I find it's useful to have the creation time in order to differentiate between similar clusters (e.g. to shut down the older one), or to make it easier to spot clusters that have been unintentionally left running.

What do you think?

### DIFF
--- a/aztk_cli/utils.py
+++ b/aztk_cli/utils.py
@@ -120,8 +120,9 @@ def print_clusters(clusters: List[models.Cluster]):
     for cluster in clusters:
         node_count = __pretty_node_count(cluster)
 
-        log.info(print_format.format(cluster.id, cluster.state.value, cluster.vm_size, node_count,
-                                     format_datetime(cluster.pool.creation_time, False)))
+        log.info(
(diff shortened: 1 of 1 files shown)


aztk_cli/utils.py
```diff
@@ -50,6 +50,7 @@ def print_cluster(client, cluster: models.Cluster, internal: bool = False):
     log.info("------------------------------------------")
     log.info("State:          %s", cluster.state.value)
     log.info("Node Size:      %s", cluster.vm_size)
+    log.info("Created:        %s", cluster.pool.creation_time.strftime('%Y-%m-%d %H:%M:%S'))
```
timotheeguerin: can you make an helper function for converting the date to string so we only provide the format once?

emlyn: Sure. I didn't include the seconds in the table listing to keep the width down, so added a parameter for that, but can remove it if you'd prefer to keep it simple.

jafreck:
@emlyn Thanks for the contribution!

PR #684, opened 2018-11-12
Clean up streaming log output

emlyn:
When streaming logs, some lines get prefixed with the local log info from aztk, while others don't, depending on whether they happen to be the first line of a batch from the remote machine. It also prints an empty log message every 3 seconds while no new logs are being generated.

I think it would be cleaner to just print the log stream without messing with it.

Or if you prefer to include the local log prefix, shouldn't the returned log messages be split into lines and logged individually so that the output is consistent?

### DIFF
--- a/aztk_cli/utils.py
+++ b/aztk_cli/utils.py
@@ -134,7 +134,8 @@ def stream_logs(client, cluster_id, application_name):
     while True:
         app_logs = client.cluster.get_application_log(
             id=cluster_id, application_name=application_name, tail=True, current_bytes=current_bytes)
-        log.print(app_logs.log)
+        if app_logs.total_bytes > current_bytes:
+            print(app_logs.log, end='', flush=True)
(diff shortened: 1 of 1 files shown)


timotheeguerin:
Good idea to check for new output before printing.
@jafreck isn't `log.print` suppose to not print anything else?

emlyn:
Ah, I must have enabled `--verbose` logging, it looks like that changes the log format.
With verbose logging enabled, the output gets a bit muddled.

timotheeguerin:
I see, we have 2 options:
* `log.print` always print the text as it is
* we split the streamoutput by new line that way each line has the prefix timestamp

jafreck:
I agree with @timotheeguerin, having `log.print` output the text without any prefix would be preferable for now. I created an issue for this: #686

jafreck:
Closing since resolution of #686  will fix this behavior in all cases where `log.print` is used.

PR #687, opened 2018-11-28
Feature: Allow apps to use Batch task dependencies

jamesclarke:
Some of the Spark applications we have in the SwiftKey team require running different steps in a pipeline where each successive step relies on the previous step completing without failure. Azure Batch jobs have a mode that allows submitting tasks that depend on one or more other tasks, only running when those other tasks have completed successfully.

I am sure the AZTK team is already aware of task dependencies and probably either made a decision not to implement full support for app task dependencies or that it was not a priority. To properly add the concept of task dependencies to AZTK itself would probably entail a lot of extra functionality around app management.

However, for our purposes, just exposing the ability to express app dependencies with the underlying Batch task option `depends_on` is enough and we are happy to implement the specific app dependency management logic we need on our side.

This pull request, therefore, is a minimal set of changes that just adds the necessary parameters/arguments to cluster job and application task creation to enable use of task dependencies in Azure Batch.

Changes:
* Turn on `use_task_dependencies` option when creating the cluster job
* Add `depends_on` argument when generating an application task
* Add `depends_on` member to `spark.models.models.ApplicationConfiguration` and corresponding parameter to its `__init__()`

### DIFF
--- a/aztk/client/cluster/helpers/create.py
+++ b/aztk/client/cluster/helpers/create.py
@@ -63,7 +63,11 @@ def create_pool_and_job_and_table(
     helpers.create_pool_if_not_exist(pool, core_cluster_operations.batch_client)
 
     # Create job
-    job = batch_models.JobAddParameter(id=job_id, pool_info=batch_models.PoolInformation(pool_id=pool_id))
+    job = batch_models.JobAddParameter(
+        id=job_id,
+        pool_info=batch_models.PoolInformation(pool_id=pool_id),
+        uses_task_depenendencies=True,
+    )
 
     # Add job to batch
     core_cluster_operations.batch_client.job.add(job)
--- a/aztk/spark/client/base/helpers/generate_application_task.py
+++ b/aztk/spark/client/base/helpers/generate_application_task.py
@@ -98,6 +98,7 @@ def generate_application_task(core_base_operations, container_id, application, r
         user_identity=batch_models.UserIdentity(
             auto_user=batch_models.AutoUserSpecification(
                 scope=batch_models.AutoUserScope.task, elevation_level=batch_models.ElevationLevel.admin)),
+        depends_on=application.depends_on,
     )
 
     return task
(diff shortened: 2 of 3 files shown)


timotheeguerin:
Ok, so I talked with Jake and there is actually a problem:
This will only work if you use Batch to submit the task not the new ssh submit mode. This means you can't use this feature with the `scheduling_target` flag. This means we need to add validation on the model to check you don't use both features.

It is also quite simple to wait for a task to finish before triggering the next one right now. Does adding this feature is really needed for your use case to make the 2 submission mode inconsistent?

We would also need to add docs for this feature

jafreck:
Achieving the functionality you are after should be possible in aztk today. Using the sdk, you can call:
```
spark_client.cluster.submit(id=cluster_id, application=application_configuration1, wait=True)

application1_log = spark_client.get_application_log(cluster_id, application_configuration1.name)
if application1_log.exit_code != 0:
    handle_error()

spark_client.cluster.submit(id=cluster_id, application=application_configuration2, wait=True)
```

I'd prefer to not utilize task dependencies on Batch for the reason Tim shared. Currently there are 2 methods for application submission. If using `scheduling_target.Master` is used, then Batch tasks are not used at all. This means that this feature would be dependent on using `scheduling_target.Any`. This sort of feature dependency is something I would like to avoid where possible. In this particular case, the above workaround seem acceptable to me if they solve your problem.

jamesclarke:
Thanks for the responses, @timotheeguerin and @jafreck. I had been working on an older version of AZTK and so, although I'd heard about the new ssh submission stuff, I haven't used it yet and thus hadn't thought of the issue you describe.

Thanks also for the suggestion about using the `wait` option whilst submitting apps sequentially. That works nicely for a number of things we're doing when we have jobs scheduled with VSTS or something and we do use that.

However, in the case I am currently working on, users need to be able to submit a batch of inter-dependent apps all at once, forming a pipeline, which may take a while to run (e.g. overnight) and so they need to be able to close their laptops and disconnect from the network. So the chain of individual submits with waiting unfortunately won't work here.

I am going to think about this for a little bit and talk to some people and then get back to this PR either with some other ideas or to just shut it down.

jamesclarke:
As per the discussion, this will not be implemented, and so closing the PR.

PR #692, opened 2019-01-04
Fix: list_applications throwing error

brettsimons:
list_applications was previously throwing this error:
```
  File "<stdin>", line 1, in <module>
  File "...\aztk\spark\client\base\operations.py", line 85, in list_applications
    return list_applications.list_applications(self, id)
  File "...\aztk\spark\client\base\helpers\list_applications.py", line 21, in list_applications
    return models.Application(_list_applications(core_operations, id))
  File "...\aztk\spark\client\base\helpers\list_applications.py", line 11, in _list_applications
    scheduling_target = core_operations.get_cluster_configuration(id).scheduling_target
AttributeError: 'ClusterOperations' object has no attribute 'get_cluster_configuration'
```

It is now working and returns a list of tasks.

### DIFF
--- a/aztk/spark/client/base/helpers/list_applications.py
+++ b/aztk/spark/client/base/helpers/list_applications.py
@@ -2,7 +2,7 @@ import azure.batch.models as batch_models
 from azure.batch.models import BatchErrorException
 
 from aztk import error
-from aztk.spark.models import SchedulingTarget
+from aztk.spark.models import Application, SchedulingTarget
 from aztk.utils import helpers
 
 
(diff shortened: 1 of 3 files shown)


brettsimons:
Here is a workaround for the current release: `Client.cluster._core_cluster_operations.list_batch_tasks(cluster_id)`

jafreck:
Thanks for reporting this, can you confirm, were you experiencing this issue on the latest released version (0.10.2)?

brettsimons:
Yes, I've only validated that this is an issue in 0.10.2. After looking a bit more at the source, I'll update the fix to return an array of aztk.spark.models.Application.

